PR #35, opened 2025-12-08
Use torchao for MXFP8 quantization

jwfromm:
Summary: Torchao offers optimized triton mxfp8 kernels that are quite a bit more efficient than the naive python implementations we currently use in our benchmarks. This diff adds torchao import and properly uses those faster quantization kernels for gemm and grouped gemm when available.

Differential Revision: D88660032


cthi:
Closing as we won't add the AO dep

PR #58, opened 2025-12-17
Add local attention mask support to CUTLASS Blackwell decode op

Aya-ZIbra:
Summary:
Extends the xformers CUTLASS Blackwell FwOpDecode operator to support BlockDiagonalLocalAttentionPaddedKeysMask, enabling sliding window attention for decode workloads. This builds on the gen kernel sliding window implementation to provide end-to-end local attention support through the xformers API.

Also for consistent shapes, I changed the default split_k_size from 1024 to 0 in the interface to use no split-K sizing.
TODO: Add merge_attentions step to cutlass op when number of splits >1

Differential Revision: D89192917


cthi:
Cleaning up stale PRs

PR #124, opened 2026-01-27
Fix AMD build

cthi:
Summary: as title, my suspicison is `HIPIFY_V2` got enabled in fbcode, and somehow this broke the build. Looking at the other CK files, I think the namespace is slightly wrong, e.g. ck_fp8_tensorwise still uses at::cuda instead of at::hip.

Differential Revision: D91586985


cthi:
Hey @jeffdaily - for `HIPIFY_V2`, should the namespace be `at::cuda::getCurrentHIPStream`? I see that in [another file](https://github.com/meta-pytorch/MSLK/blob/2d63d106bbbe2876feef0a1c8d662511ba3b6de6/csrc/quantize/ck/ck_utility.hip#L27-L39C29) that I think you updated in the past, it is using `at::cuda::` instead of `at::hip::`. Thanks!

jeffdaily:
HIPIFY_V2 should be using c10::cuda::getCurrentCUDAStream.  The point of v2 is that it doesn't rename classes or namespaces from cuda->hip.

jeffdaily:
FYI the hipify v2 PR was reverted.  Working on fixing it.

cthi:
Closing in favor of https://github.com/meta-pytorch/MSLK/pull/131

PR #133, opened 2026-02-05
fp8 quantize rowwise speedup

coreyhu:
Summary:
Rowwise Quantization:
  Average Speedup:       1.08x
  Median Speedup:        1.04x
  Max Speedup:           1.41x
  Min Speedup:           0.98x
  Win Rate:             94.3% (33/35 cases)
  Avg Throughput Gain:   +193.0 GB/s

Differential Revision: D91731095


cthi:
Cleaning up stale PRs

PR #154, opened 2026-02-16
fix unused var

bottler:
Summary: No lse from FwOpDecode.

Differential Revision: D93427253


cthi:
Cleaning up stale PRs

PR #157, opened 2026-02-18
Revert D93220414

bottler:
Summary:
This diff reverts D93220414
seems to break mslk/csrc/attention/ck/fmha:fmha build on dev-nosan-amd-gpu because of linker overflow. Which breaks tests. see https://www.internalfb.com/sandcastle/workflow/1756403854688273030 https://fburl.com/scuba/testinfra_db_results/g8e8fqay

Depends on D93220414

Reviewed By: bottler

Differential Revision: D93669401


cthi:
Cleaning up stale PRs

PR #187, opened 2026-02-26
CuteDSL BF16 Grouped GEMM

jiawenliu64:
Summary:
Build CuteDSL BF16 grouped GEMM kernel on blackwell

Performance-wise, the new CuteDSL BF16 grouped GEMM kernel is comparable to torch BF16 grouped GEMM, while I will further improve its performance to make it faster across all shapes

Accuracy-wise, its bit parity is the same as torch kernel

Differential Revision: D94540669


cthi:
Closing old PRs

PR #237, opened 2026-03-20
Fix CK fmha

cthi:
Summary: I believe there was a very subtle bug when the cpp_lib.py got cleaned up in D96328593. The `cpp_lib.py` is responsbiel to load the `.so` (see `_register_extensions`) for CK, but an import of it was removed in `common.py`. As such the torch op registrations for CK would not be loaded as the python module code would not run due to lack of any import.

Differential Revision: D97407414


bottler (approved):

PR #342, opened 2026-05-12
Switch to upstream Cutlass dependency

jwfromm:
Summary: The issues that motivated the use of a custom cutlass copy no longer exist in upstream cutlass. This diff finally moves our dependency over to standard Cutlass 4.42. In OSS, we similarly will drop the use of `https://github.com/jwfromm/cutlass`.

Differential Revision: D104857383


cthi:
Closing old PRs

PR #351, opened 2026-05-21
remove unused file and add script

cthi:
Differential Revision: D105957735


bottler (approved):

PR #375, opened 2026-06-08
[ROCm] Add Triton MXFP8×MXFP4 GEMM kernel for ROCm gfx950 (CDNA4)

avbokovoy:
## Summary

This PR adds Triton backend for `mx8mx4bf16` / `mx8mx4bf16_grouped_mm` operations on ROCm, enabling MX8xMX4 tests and benchmarks.

- Rename `CutlassMX8MX4Groupwise` and `CutlassMX8MX4GroupwiseGrouped2D3D`
  benchmark ops to `MX8MX4Groupwise` and `MX8MX4GroupwiseGrouped2D3D`; add
  AMD_GFX950 to their supported_accelerators so they run in gemm_bench
  on MI350X
- Add AMD_GFX950 Accelerator enum value and gfx950 detection logic in
  get_current_accelerator() using gcnArchName (also tightens gfx942
  detection for MI300X)
- Move `mx8mx4bf16` and `mx8mx4bf16_grouped_mm` op schema registrations
  above the USE_ROCM guard in `gemm_ops.cpp` so the schemas are shared
  between CUDA and ROCm builds
- Register Triton `mx8mx4bf16` / `mx8mx4bf16_grouped_mm` implementations
  at package import time via `mslk/gemm/__init__.py` on ROCm builds
- Move grouped MX8xMX4 GEMM tests out of MXFP4Tests and into
  `MX8MX4Tests`; lift the NVIDIA-only skip condition so the tests run
  on gfx950; import _to_blocked unconditionally

## Test plan

- [x] MX8MX4Tests::test_mx8mx4_grouped_gemm_2d_3d - passes on ROCm
- [x] MX8MX4Tests::test_mx8mx4_grouped_gemm_2d_3d_empty_groups - passes on ROCm
- [x] MX8MX4Tests::test_gemm - passes on ROCm
- [x] Flake8 linter passes
- [x] No regressions on unrelated test classes

### DIFF
--- a/bench/gemm/gemm_ops.py
+++ b/bench/gemm/gemm_ops.py
@@ -3080,18 +3080,19 @@ class MX8MX4GroupwiseGrouped2D3D(GemmOpBase):
     def preprocess(self, x, w):
         assert isinstance(x, list)
         assert isinstance(w, list)
+        # Cache per-group M sizes before concatenation so quantize can use
+        # the exact boundaries regardless of whether groups are equal-sized.
+        self._m_sizes = [xi.shape[0] for xi in x]
         x = torch.cat(x, dim=0).contiguous()
         w = torch.stack(w, dim=0).contiguous()
         return x, w
 
     def quantize(self, x, w):
-        block_size = 32
         G, _, K = w.shape
-        total_M = x.shape[0]
-        group_size = total_M // G
-        input_group_end_offsets = torch.arange(
-            group_size, total_M + 1, group_size, dtype=torch.int32, device=x.device
-        )
+        m_sizes = self._m_sizes
+        input_group_end_offsets = torch.tensor(
+            m_sizes, dtype=torch.int32, device=x.device
+        ).cumsum(0)
 
         wq_list = []
         w_scale_list = []
@@ -3104,19 +3105,18 @@ class MX8MX4GroupwiseGrouped2D3D(GemmOpBase):
 
         xq_list = []
         x_scale_list = []
-        for i in range(G):
-            prev_group_end = 0 if i == 0 else input_group_end_offsets[i - 1]
-            curr_group_end = input_group_end_offsets[i]
-            x_slice = x[prev_group_end:curr_group_end, :]
+        row = 0
+        for m_g in m_sizes:
+            x_slice = x[row : row + m_g, :]
             x_scale, xq = to_mxfp8(x_slice)
-            x_scale = _to_blocked(
-                x_scale.view(torch.int8).reshape(x_slice.shape[0], -1)
-            ).view(torch.uint8)
+            x_scale = _to_blocked(x_scale.view(torch.int8).reshape(m_g, -1)).view(
+                torch.uint8
+            )
             xq_list.append(xq)
             x_scale_list.append(x_scale)
+            row += m_g
         xq = torch.cat(xq_list, dim=0).contiguous()
         x_scale = torch.cat(x_scale_list, dim=0).contiguous()
-        x_scale = x_scale.reshape(-1, K // block_size)
         return xq, wq, x_scale, w_scale, input_group_end_offsets
 
     def compute(self, xq, wq, x_scale, w_scale, input_group_end_offsets):
--- a/bench/gemm/gemm_ops.py
+++ b/bench/gemm/gemm_ops.py
@@ -31,7 +31,6 @@ from mslk.quantize.triton.fp8_quantize import (
 )
 from mslk.utils.triton.fp8_utils import get_fp8_constants
 
-
 try:
     from tinygemm.utils import group_quantize_tensor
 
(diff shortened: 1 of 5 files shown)


cthi:
> Important note: MX8MX4GroupwiseGrouped2D3D only supports --no-cuda-graph launch only on ROCm for now.

Oh do we have some D2H op in the benchmarking?

avbokovoy:
> > Important note: MX8MX4GroupwiseGrouped2D3D only supports --no-cuda-graph launch only on ROCm for now.
> 
> Oh do we have some D2H op in the benchmarking?

@cthi  It looks like, but I didn't dig much into it tbh. For example this command: `python bench/gemm/gemm_bench.py   --kernels MX8MX4GroupwiseGrouped2D3D   --grouped   --M 512,256,128,64   --N 7168,7168,7168,7168   --K 3584,3584,3584,3584` fails with:
```bash
torch.AcceleratorError: HIP error: operation not permitted when stream is capturing
Search for `hipErrorStreamCaptureUnsupported' in https://rocm.docs.amd.com/projects/HIP/en/latest/index.html for more information.
HIP kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing AMD_SERIALIZE_KERNEL=3
Compile with `TORCH_USE_HIP_DSA` to enable device-side assertions.
```

Adding `--no-cuda-graph` flag makes benchmark work. Let me know if you want me to look into this issue

avbokovoy:
Benchmark should work without `--no-cuda-graph` flag

mslk/gemm/triton/mx8mx4_gemm.py
```diff
@@ -0,0 +1,591 @@
...
+MXFP8 (E4M3) activation x MXFP4 (E2M1) weight GEMM for ROCm/AMD gfx950+.
+
+This is the ROCm counterpart of csrc/gemm/cutlass/mx8mx4bf16.cu.
+It uses Triton's tl.dot_scaled, which lowers to the CDNA4 scaled-MFMA
+instructions (V_MFMA_SCALE_F32_16X16X128_F8F6F4) available on gfx950
+(MI350/MI355X).  The kernel is NOT compatible with gfx942 (MI300X) or
+older CDNA architectures which lack hardware block-scaling support.
+
+Tensor shapes (matching the CUDA CUTLASS convention):
+  XQ      : [M, K]     — MXFP8 E4M3, stored as torch.float8_e4m3fnuz
```
cthi: nit: I think the fp8 dtype is the ocp variant right? not fnuz

bench/gemm/gemm_ops.py
```diff
@@ -3064,26 +3072,27 @@ def compute_dtype(self) -> ComputeDtype:
...
         G, _, K = w.shape
-        total_M = x.shape[0]
-        group_size = total_M // G
-        input_group_end_offsets = torch.arange(
-            group_size, total_M + 1, group_size, dtype=torch.int32, device=x.device
-        )
+        m_sizes = self._m_sizes
+        input_group_end_offsets = torch.tensor(
+            m_sizes, dtype=torch.int32, device=x.device
+        ).cumsum(0)
```
cthi: I think you need to ensure the `cumsum` is kept as `int32`? As it would be `int64` by default. I think the nvidia kernel may not cast it.

test/gemm/gemm_test.py
```diff
@@ -2914,9 +2794,11 @@ def test_mxfp4_16_vs_nvfp4_scale_range(self) -> None:
         )
 
 
-@unittest.skipIf(not SUPPORTS_MXFP4, "Skip if MXFP4 is not supported")
+@unittest.skipIf(
+    not SUPPORTS_MXFP4 and not torch.version.hip, "Skip if MXFP4 is not supported"
```
cthi: On rocm, this test should only run for gfx950 right?

cthi (commented):
Looks good, left a few small comments.

test/gemm/gemm_test.py
avbokovoy: Right. Fixed in https://github.com/meta-pytorch/MSLK/pull/375/commits/5edc9ff0aae6b89fc6305f7a9dedfe6fc8c3a5f3

bench/gemm/gemm_ops.py
avbokovoy: Explicitly specified int32 type for cumsum in https://github.com/meta-pytorch/MSLK/pull/375/commits/8a314b0186f7c9a03f00d8bf090cb6196273ea8f

mslk/gemm/triton/mx8mx4_gemm.py
avbokovoy: OCP it is. Fixed docs in https://github.com/meta-pytorch/MSLK/pull/375/commits/313b4c59a91f148de614f5f151ff0397be20cbb2

PR #377, opened 2026-06-09
[ROCm] Add Triton MLA decode + prefill kernels for MI300X

kudomcho:
## Motivation

Add Triton MLA (Multi-head Latent Attention) decode + prefill kernels for AMD MI300X (gfx942) to enable DeepSeek-V3 inference on ROCm. MLA uses weight-absorbed KV cache (NUM_HEADS=128, NUM_KV_HEADS=1, KV_LORA_RANK=512, QK_ROPE_HEAD_DIM=64) where K and V share storage, requiring specialized attention kernels distinct from standard GQA/MQA.

Decode kernels ported from SGLang (`sgl-project/sglang`, `triton_ops/decode_attention.py`, weight absorption variant with `HAS_MLA=True`). Prefill kernel is original work (SGLang has no MLA-specific Triton prefill).

## Technical Details

**`mslk/attention/mla/_triton/mla_kernels.py`** — Triton JIT kernels + launchers:
- **Decode** (`_fwd_grouped_kernel_stage1` + `_fwd_kernel_stage2`): ported from SGLang's split-K grouped-head pipeline. Grid: `(batch, head_groups, MAX_KV_SPLITS)`. SGLang patterns preserved: `BLOCK_DMODEL`/`BLOCK_DPE`/`BLOCK_DV` naming, `MIN_BLOCK_KV=32` alignment, HIP-specific `waves_per_eu`/`kpack` kwargs, `v_scale` for FP8 dequant. MSLK additions: direct block_tables indexing (no kv_indptr/kv_indices), fused single-kernel for splits=1, 4-stage pipelining.
- **Prefill** (`_mla_prefill_fwd`): single-stage flash attention with causal masking, per-head-group tiling (BLOCK_H=64), token-level block_tables indexing, BLOCK_Q tiling support with per-token sequence lookup.
- FP8 (e4m3fnuz) and BF16 inputs, BF16 output, FP32 accumulation. Paged KV cache.

**`mslk/attention/mla/triton_mla.py`** — Public API:
- `mla_decode_fwd()` / `mla_prefill_fwd()`: arch-aware tuning (gfx942 vs gfx950), adaptive split-K selection

**`mslk/attention/mla/__init__.py`** — re-exports `mla_decode_fwd` and `mla_prefill_fwd`

**`test/attention/test_mla.py`** — 43 correctness tests (cosine distance < 3e-2 vs PyTorch reference)

MLA is placed in its own `mslk/attention/mla/` module (not inside `fmha/`) since it has a fundamentally different input format from standard attention (single shared KV head, no separate K/V tensors). Pure Python/Triton — no C++ op registration needed.

## Test Plan

```
XFORMERS_ENABLE_TRITON=1 pytest test/attention/test_mla.py -v
```

43 tests covering:
- Decode: BF16 + FP8, batch ∈ {1,3,7,13,16,64,128,256}, ctx ∈ {64..8192}, variable seqlens, non-aligned ctx, output shape, reproducibility
- Prefill: BF16 + FP8, batch ∈ {1,2,3,7}, ctx ∈ {64..4096}, chunked prefill (qlen < ctx_len)
- Cross-kernel: decode-prefill consistency for qlen=1

## Test Result

All 43 correctness tests passed on MI300X (gfx942). All tests gated on `torch.version.hip` — NVIDIA CI unaffected.

### Decode performance (qlen=1, page_size=1, BF16)

| batch | ctx_len | latency (ms) | GB/s    |
|-------|---------|-------------|---------|
| 1     | 512     | 0.123       | 24.2    |
| 1     | 4096    | 0.125       | 56.7    |
| 8     | 512     | 0.119       | 199.6   |
| 8     | 4096    | 0.117       | 484.9   |
| 16    | 1024    | 0.112       | 507.9   |
| 16    | 4096    | 0.124       | 917.8   |
| 64    | 1024    | 0.140       | 1,144.5 |
| 64    | 4096    | 0.385       | 1,006.3 |
| 128   | 2048    | 0.394       | 1,027.5 |
| 128   | 4096    | 0.717       | 986.6   |
| 256   | 512     | 0.310       | 1,151.5 |
| 256   | 2048    | 0.860       | 941.9   |

**Decode analysis:** Memory-bandwidth bound. Measured achievable peaks on this MI300X: 3,648 GB/s (sequential HBM read), 2,858 GB/s (GEMV [128×524K]). Our peak of 1,151 GB/s = **40% of GEMV ceiling**. The remaining gap is the online softmax dependency chain (max → exp → rescale → accumulate) between KV tile loads, which prevents full load/compute overlap. This is fundamental to online attention and cannot be optimized within the current algorithm. 4-stage pipelining already hides as much latency as the register budget allows.

### Prefill performance (page_size=16, BF16)

| batch | ctx_len | latency (ms) | TFLOPS |
|-------|---------|-------------|--------|
| 1     | 512     | 0.279       | 262.0  |
| 1     | 1024    | 0.858       | 340.4  |
| 1     | 2048    | 2.979       | 392.2  |
| 1     | 4096    | 11.172      | 418.3  |
| 2     | 512     | 0.511       | 286.0  |
| 2     | 1024    | 1.645       | 355.2  |
| 2     | 2048    | 5.829       | 400.9  |
| 4     | 512     | 0.978       | 298.5  |
| 4     | 1024    | 3.218       | 363.0  |

**Prefill analysis:** Compute-bound. The MI300X BF16 spec is 1,307 TFLOPS, but even cuBLAS GEMM [8K×8K] only achieves 634 TFLOPS (49% of spec). Our peak of 418 TFLOPS = **66% of achievable GEMM peak**. The remaining gap comes from: (1) online softmax overhead (exp, max, rescale) between dot products, (2) thin tile dimensions — MLA's per-tile dots are (64×512)×(512×16), where N=16 underutilizes MFMA, and (3) the (64×512) FP32 accumulator is 128KB, exceeding MI300X's 64KB LDS and spilling to registers. Optimizations attempted: BLOCK_Q multi-token tiling (sequential loop overhead negates KV reuse), BF16 accumulators (precision loss in softmax rescaling), pipeline stages 2-4 (register spill from accumulator), TILE_SIZE=32+ (LDS overflow). All were measured to be neutral or negative.

### Key MI300X tuning (gfx942, 64KB LDS)

**Decode:** adaptive splits (1-8), BN=16, BH=64, warps=4, stages=4. Direct block_tables indexing. Fused single-kernel for splits=1. Weight absorption: `V = transpose(K[:BLOCK_DMODEL])`.

**Prefill:** BH=64, TILE_SIZE=16, BLOCK_Q=1, warps=4, stages=1. Token-level block_tables indexing.

### Further optimization path

Both kernels are near their practical MI300X ceilings. Further gains require:
- **gfx950 (MI350X)**: 128KB LDS would enable TILE_SIZE=32 and pipeline stages for prefill; tuning configs are included but not yet validated
- **Algorithmic change**: ring attention or non-online-softmax formulations (trades numerical stability for throughput)
- **CK/HIP kernels**: hand-tuned assembly for tighter MFMA scheduling, bypassing Triton's register allocator constraints

### DIFF
--- a/csrc/attention/ck/fmha/attention.cpp
+++ b/csrc/attention/ck/fmha/attention.cpp
@@ -31,11 +31,5 @@ TORCH_LIBRARY_FRAGMENT(xformers, m) {
   m.def(TORCH_SELECTIVE_SCHEMA(
       "xformers::efficient_attention_forward_decoder_splitk_ck(Tensor query, Tensor key, "
       " Tensor value, Tensor? seq_positions, float scale, int split_k) -> Tensor"));
-  m.def(TORCH_SELECTIVE_SCHEMA(
-      "xformers::efficient_attention_mla_decode(Tensor query, Tensor kv_buffer, "
-      "Tensor block_tables, Tensor cu_seqlens_q, Tensor seqused_k, float scale) -> Tensor"));
-  m.def(TORCH_SELECTIVE_SCHEMA(
-      "xformers::efficient_attention_mla_prefill(Tensor query, Tensor kv_buffer, "
-      "Tensor block_tables, Tensor cu_seqlens_q, Tensor seqused_k, float scale) -> Tensor"));
 #endif
 }
--- a/csrc/attention/ck/fmha/attention.cpp
+++ b/csrc/attention/ck/fmha/attention.cpp
@@ -31,5 +31,11 @@ TORCH_LIBRARY_FRAGMENT(xformers, m) {
   m.def(TORCH_SELECTIVE_SCHEMA(
       "xformers::efficient_attention_forward_decoder_splitk_ck(Tensor query, Tensor key, "
       " Tensor value, Tensor? seq_positions, float scale, int split_k) -> Tensor"));
+  m.def(TORCH_SELECTIVE_SCHEMA(
+      "xformers::efficient_attention_mla_decode(Tensor query, Tensor kv_buffer, "
+      "Tensor block_tables, Tensor cu_seqlens_q, Tensor seqused_k, float scale) -> Tensor"));
+  m.def(TORCH_SELECTIVE_SCHEMA(
+      "xformers::efficient_attention_mla_prefill(Tensor query, Tensor kv_buffer, "
+      "Tensor block_tables, Tensor cu_seqlens_q, Tensor seqused_k, float scale) -> Tensor"));
 #endif
 }
--- a/mslk/attention/fmha/__init__.py
+++ b/mslk/attention/fmha/__init__.py
@@ -9,7 +9,7 @@ from typing import Any, cast, List, Optional, Sequence, Tuple, Type, Union
 
 import torch
 
-from . import (
+from . import (  # noqa: F401
     attn_bias,
     ck,
     ck_decoder,
@@ -21,6 +21,7 @@ from . import (
     flash,
     flash3,
     flash_mtia,
+    triton_mla,
     triton_splitk,
 )
 from .attn_bias import (
@@ -148,9 +149,9 @@ class _fMHA(torch.autograd.Function):
         ):
             varlen_lse_packed = _detect_lse_packed_or_raise(op_ctx.lse, inp)
             if varlen_lse_packed is not None and op_fw is not None:
-                assert op_fw.VARLEN_LSE_PACKED == varlen_lse_packed, (
-                    f"{op_fw.NAME}: wrong value for `VARLEN_LSE_PACKED` ?"
-                )
+                assert (
+                    op_fw.VARLEN_LSE_PACKED == varlen_lse_packed
+                ), f"{op_fw.NAME}: wrong value for `VARLEN_LSE_PACKED` ?"
             # NOTE: We need to check tensor strides to decide which operator we run in the BW pass.
             # Unfortunately, PyTorch only allows to call this function during the FW pass, so
             # we decide the operator to use now.
@@ -959,6 +960,7 @@ ALL_FW_OPS: List[Type[AttentionFwOpBase]] = [
     flash_mtia.FwOp,
     flash3.FwOp,
     triton_splitk.FwOp,
+    triton_mla.FwOp,
 ]
 
 ALL_BW_OPS: List[Type[AttentionBwOpBase]] = [
--- a/mslk/attention/fmha/__init__.py
+++ b/mslk/attention/fmha/__init__.py
@@ -21,7 +21,6 @@ from . import (  # noqa: F401
     flash,
     flash3,
     flash_mtia,
-    triton_mla,
     triton_splitk,
 )
 from .attn_bias import (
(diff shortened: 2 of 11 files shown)


csrc/attention/ck/fmha/attention.cpp
```diff
@@ -31,5 +31,11 @@ TORCH_LIBRARY_FRAGMENT(xformers, m) {
   m.def(TORCH_SELECTIVE_SCHEMA(
       "xformers::efficient_attention_forward_decoder_splitk_ck(Tensor query, Tensor key, "
       " Tensor value, Tensor? seq_positions, float scale, int split_k) -> Tensor"));
+  m.def(TORCH_SELECTIVE_SCHEMA(
```
cthi: I think we would not need to add C++ op registration, we could leave it as Python?

mslk/attention/fmha/triton_mla.py
```diff
@@ -0,0 +1,263 @@
...
+            reasons.append(
+                f"MLA requires qk_head_dim={MLA_QK_HEAD_DIM}, got {d.query.shape[-1]}"
+            )
+        return reasons
+
+    @classmethod
+    def apply(
+        cls, inp: Inputs, needs_gradient: bool
+    ) -> Tuple[torch.Tensor, Optional[Context]]:
+        raise NotImplementedError(
```
cthi: Since MLA is a bit different from standard fMHA, does it make sense for us to move this outside of fmha? e.g. Place in `mslk/attention/mla` ? Then you won't need to worry about these fmha specific things.

cthi (commented):
Thanks for the contribution, it looks good. Only comment is we should probably place this in `mslk/attention/mla` instead of `mslk/attention/fmha`, as this is a bit different fmha and then you wouldn't need to worry about the fmha specific things, since the caller would just call it through a different path anyways.

csrc/attention/ck/fmha/attention.cpp
kudomcho: Done — removed the C++ op schemas and kept MLA as pure Python/Triton. Also moved everything to `mslk/attention/mla/` per your other comment.

mslk/attention/fmha/triton_mla.py
kudomcho: Good call — moved to `mslk/attention/mla/` with its own `__init__.py` that exports `mla_decode_fwd` and `mla_prefill_fwd`. Removed the `FwOp` class and all fmha dispatch integration. All 43 tests pass from the new path.

mslk/attention/fmha/__init__.py
```diff
@@ -9,7 +9,7 @@
 
 import torch
 
-from . import (
+from . import (  # noqa: F401
```
cthi: I think you forgot to revert the change to this file

kudomcho: reverted both `fmha/__init__.py` and `fmha/dispatch.py` to match main. They had leftover formatting changes from the earlier commits.

bottler (approved):

PR #381, opened 2026-06-10
Add Triton ROCm backend for MXFP4 GEMM ops MI350X

amd-weisun:
## Summary

Adds AMD ROCm backends for the three FP4 GEMM ops that previously existed only
as CUDA/CUTLASS (SM100): `f4f4bf16`, `f4f4bf16_grouped_mm`, and
`f4f4bf16_grouped_stacked`. The implementation is pure Triton with
**native MXFP4 only** on **gfx950 (MI350X / MI355X)** via `tl.dot_scaled`
scaled-MFMA.

**ROCm scope (intentionally narrow):**

- **Supported:** MXFP4 block-size 32 (E8M0 scales) on gfx950 — dense, grouped
  2D-3D (MoE), and grouped-stacked layouts.
- **Not supported on ROCm:** NVFP4, MXFP4_16, gfx942 (MI300X) emulation
  (dequant→FP8), and CUDA-only grouped 2D-2D (K-grouped) layout.

Emulation paths (dequant→FP8 for NVFP4 or MXFP4 on hardware without native FP4)
were prototyped but dropped: they are not numerically equivalent to native scaled
FP4 MFMA and would confuse users about expected accuracy. Use CUDA for NVFP4 /
MXFP4_16; use gfx950 for native MXFP4.

## Background

The three target ops map 1:1 to the existing CUDA kernels in
`csrc/gemm/cutlass/f4f4bf16*.cu` and keep their exact op signatures, so the
Python frontend dispatches to either CUDA (CUTLASS/SM100) or ROCm (this Triton
path) by device with no API change.

| Format | Block | Element | Scale | Global | ROCm |
|---|--:|---|---|---|:---:|
| MXFP4 (OCP MX v1.0) | 32 | E2M1 | UE8M0 | — | gfx950 native |
| MXFP4_16 | 16 | E2M1 | UE8M0 | — | CUDA only |
| NVFP4 (Blackwell) | 16 | E2M1 | UE4M3 | per-tensor FP32 | CUDA only |

gfx950 (MI350X) has native MXFP4 scaled-MFMA (`v_mfma_scale_f32_*_f8f6f4`).
gfx942 (MI300X) has no native FP4 hardware; this PR does not provide a fallback
there.

## Implementation

All FP4 Triton code lives in `mslk/gemm/triton/f4f4bf16.py`, following the
`fp8_gemm.py` "one dtype family per file" precedent. Backend selection is
inline (`torch.version.hip` / `gcnArchName`), matching the existing convention.

**Kernels**

| Kernel | Purpose |
|---|---|
| `_mxfp4_gemm_kernel` | Native MXFP4 via `tl.dot_scaled` (gfx950). Also backs both grouped ops. |
| `_mxfp4_grouped_mm_kernel` | Grouped MXFP4 (2D-3D MoE layout); shared by `grouped_mm` and `grouped_stacked` via stride remap. |

**Quantizer** (`mslk/quantize/triton/fp4_quantize.py`): AMD kernels consume a
plain `[M, K//32]` scale layout (not the TCGen5-swizzled Blackwell one), so
`triton_quantize_mx4_unpack` gains an `IS_ROCM` branch that emits the plain
layout on ROCm.

**Op registration** (`mslk/gemm/_meta.py`): ROCm-only `torch.library` fake-impl
registrations for the three ops, so `torch.compile` / meta tracing works.

## Path matrix

| Device | Input | Path |
|---|---|---|
| MI350X / MI355X (gfx950) | MXFP4 dense / grouped 2D-3D / grouped-stacked | native scaled MFMA |
| MI350X | NVFP4, MXFP4_16, grouped 2D-2D | not supported (use CUDA) |
| MI300X (gfx942) | any FP4 GEMM | not supported |

## Performance

Measured on **MI350X (gfx950)**, torch 2.10 + ROCm 7.2.4 + Triton 3.6, over
Llama-3-8B-class projection shapes (o N=4096 K=4096, qkv N=6144, gate/up
N=14336, down K=14336) plus square ceiling reads. Numbers are TFLOPS
(`2·M·N·K / time`, best-of autotune, steady state); correctness is covered
separately by the unittests. The same ops are registered in
`bench/gemm/gemm_ops.py` for the standard harness.

**mx_native** — MXFP4 via native `tl.dot_scaled` scaled-MFMA (the only ROCm path).

### MI350X — decode / small-M (TFLOPS)

| M | N | K | mx_native |
|--:|--:|--:|--:|
| 16 | 14336 | 4096 | 63 |
| 64 | 14336 | 4096 | 264 |
| 128 | 14336 | 4096 | 524 |
| 256 | 14336 | 4096 | 1051 |
| 128 | 4096 | 4096 | 141 |

Small-M is launch/latency-bound; decode-phase callers should wrap the op in a
HIP graph to elide launcher overhead (standard MSLK decode pattern).

### MI350X — prefill / large-M (TFLOPS)

| M | N | K | mx_native |
|--:|--:|--:|--:|
| 512 | 14336 | 4096 | 1680 |
| 1024 | 14336 | 4096 | 2458 |
| 2048 | 14336 | 4096 | 2514 |
| 4096 | 4096 | 14336 | 3256 |
| 8192 | 14336 | 4096 | 2794 |
| 8192 | 4096 | 14336 | 3305 |

Native MXFP4 peaks ~3300 TFLOPS (~55% of the MI350X MXFP4 MFMA ceiling).
Post-scope-trim validation on aac15 (4096³ dense) reports ~2880–2940 TFLOPS —
unchanged vs the pre-trim native path (emulation code was dead on gfx950).

## Accuracy

Correctness is asserted by the unittests: each op's output is compared against
a `bf16` reference matmul with `torch.testing.assert_close`, using the same
shapes and tolerances as the existing CUDA tests (MXFP4 `atol/rtol 8e-2`).

Native scaled MFMA on gfx950 is the only ROCm path; there is no separate
emulation accuracy story on AMD.

## Testing

**CUDA path unchanged by construction.** Every change in this PR is AMD-scoped.

`test/gemm/gemm_test.py`, arch-aware capability gates:

- `supports_mxfp4()` / `supports_mxfp4_native()` → gfx950 (and CUDA SM100).
- `supports_nvfp4()` → CUDA SM100 only (no ROCm NVFP4).

| Test | MI350X (gfx950) | MI300X (gfx942) |
|---|:--:|:--:|
| `MXFP4Tests::test_gemm` (dense) | PASS | SKIP |
| `MXFP4Tests::test_grouped_gemm_2d_3d` | PASS | SKIP |
| `MXFP4Tests::test_grouped_gemm_2d_2d` | SKIP (CUDA-only 2D-2D layout) | SKIP |
| `MXFP4Tests::test_grouped_stacked_gemm` | PASS | SKIP |
| `MXFP4Tests::test_grouped_stacked_gemm_uneven` | PASS | SKIP |
| `NVFP4Tests::*` | SKIP (CUDA only) | SKIP |
| `MXFP4BlockSize16Tests::*` | SKIP (CUDA only) | SKIP |

`test_grouped_stacked_gemm_uneven` uses uneven expert sizes as an edge-case
regression guard for the grouped launch-grid sizing.

## Files changed

| File | Change |
|---|---|
| `mslk/gemm/triton/f4f4bf16.py` | new — native MXFP4 kernels, wrappers, dispatcher |
| `mslk/quantize/triton/fp4_quantize.py` | `IS_ROCM` plain-scale-layout branch in `triton_quantize_mx4_unpack` |
| `mslk/gemm/_meta.py` | +ROCm `torch.library` registrations for the 3 ops |
| `test/gemm/gemm_test.py` | arch gates; ROCm skips for CUDA-only layouts/formats |
| `bench/gemm/gemm_ops.py` | `Accelerator.AMD_MI350X` + 3 MXFP4 `GemmOpBase` classes |

## Test plan

Tested on **MI350X (gfx950)**, torch + ROCm 7.2 + Triton 3.6:

- [x] `MXFP4Tests::test_gemm` (dense, native MXFP4) — pass
- [x] `MXFP4Tests::test_grouped_gemm_2d_3d` — pass
- [x] `MXFP4Tests::test_grouped_stacked_gemm` — pass
- [x] `MXFP4Tests::test_grouped_stacked_gemm_uneven` — pass
- [x] `MXFP4Tests::test_grouped_gemm_2d_2d` — skip (CUDA-only layout)
- [x] NVFP4 / MXFP4_16 tests — skip on ROCm

### DIFF
--- a/mslk/gemm/triton/f4f4bf16.py
+++ b/mslk/gemm/triton/f4f4bf16.py
@@ -443,9 +443,16 @@ def mxfp4_grouped_mm(
         x_scale = x_scale.view(torch.uint8)
     if w_scale.dtype != torch.uint8:
         w_scale = w_scale.view(torch.uint8)
-    assert offsets.dtype == torch.int32 and offsets.dim() == 1
+        assert offsets.dtype == torch.int32 and offsets.dim() == 1
 
-    total_M = XQ.shape[0]
+        if WQ.dim() != 3:
+            raise RuntimeError(
+                "ROCm f4f4bf16_grouped_mm only supports 2D-3D grouped GEMM "
+                "(XQ [total_M, K//2], WQ [G, K//2, N] after caller transpose). "
+                "2D-2D K-grouped layout is CUDA-only."
+            )
+
+        total_M = XQ.shape[0]
     G = WQ.shape[0]
     # WQ here is [G, K//2, N] (already transposed by caller per the CUTLASS
     # convention WQ.transpose(-2,-1).is_contiguous()).
--- a/mslk/gemm/triton/f4f4bf16.py
+++ b/mslk/gemm/triton/f4f4bf16.py
@@ -8,24 +8,13 @@
 """
 Triton f4f4bf16 GEMM for AMD ROCm (gfx950 / MI350x).
 
-Two paths in one file, both consuming the same `f4f4bf16` op signature used by
-the CUTLASS SM100 kernel in ``csrc/gemm/cutlass/f4f4bf16.cu``:
+Native MXFP4 only: ``mxfp4_gemm`` uses ``tl.dot_scaled`` with
+``lhs_format="e2m1"``, ``rhs_format="e2m1"``, ``scale_format="e8m0"``,
+block_size=32 on gfx950 (MI350x) ``v_mfma_scale_f32_*_f8f6f4``.
 
-- ``mxfp4_gemm``: native MXFP4 × MXFP4 → BF16 using ``tl.dot_scaled`` with
-  ``lhs_format="e2m1"``, ``rhs_format="e2m1"``, ``scale_format="e8m0"``,
-  block_size=32. Targets gfx950 (MI350x) ``v_mfma_scale_f32_*_f8f6f4``.
-
-- ``nvfp4_dequant_gemm``: NVFP4 (block_size=16, E4M3 scales + per-tensor FP32
-  global_scale) → dequant to arch-native fp8 (block scale folded in) once into
-  HBM → tuned fp8 MFMA GEMM. NVFP4's 16-element E4M3 block scales are not
-  compatible with the native MXFP4 scaled MFMA (32-element E8M0), so NVFP4 uses
-  this path on both gfx942 (MI300X) and gfx950 (MI350X).
-
-A third entry point, ``mxfp4_dequant_gemm``, runs the MXFP4 path on hardware
-without native FP4 MFMA (gfx942) through the same pre-dequant→fp8 path. Both
-share ``_fp4_dequant_to_fp8_kernel`` (switched by an ``IS_NVFP4`` flag) feeding
-an arch-native fp8 GEMM — converting each FP4 element exactly once rather than
-re-dequantizing per output tile.
+NVFP4 and MXFP4 emulation (dequant→FP8 fallback on gfx942 / MI300X) are
+intentionally not implemented on ROCm — use CUDA for NVFP4/MXFP4_16, and gfx950
+for native MXFP4 block_size=32.
 """
 
 from typing import Optional
@@ -190,13 +179,8 @@ def mxfp4_gemm(
     w_scale: torch.Tensor,    # [N, K//32] uint8 (E8M0)
     output: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
-    """MXFP4 × MXFP4 → BF16.
-
-    On gfx950 / SM100 uses native ``tl.dot_scaled`` MFMA. On gfx942 (no native
-    FP4 hardware) falls back to the pre-dequant→fp8 path.
-    """
-    if not _has_native_mxfp4_mfma():
-        return mxfp4_dequant_gemm(XQ, WQ, x_scale, w_scale, output=output)
+    """MXFP4 × MXFP4 → BF16 via native ``tl.dot_scaled`` MFMA (gfx950 only on ROCm)."""
+    _require_native_mxfp4()
     # Accept either raw uint8 or torch's packed FP4 dtypes (float4_e2m1fn_x2 for
     # the packed nibble pair, float8_e8m0fnu for the E8M0 scale). The kernel
     # only knows uint8; view the typed tensors as uint8 for indexing arithmetic.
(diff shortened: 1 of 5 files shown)


bottler (approved):

cthi:
Thanks for this contribution, I imported it internally and rebased it. The native MXFP4 looks great.

For the emulated paths (for MXFP4 and NVFP4) I have a few questions:
- It looks for these paths where we don't have native FP4 support we would leverage dequant the quantized FP4 value into FP8, and leverage FP8 gemm. This would not be equivalent to native FP4 since theres no guarantee the FP8 values could hold the (fp4_quant * scale). Curious if we should document this a bit more, since users could get confused about the different numerical behavior and perhaps leverage a different API for it (e.g. mxfp4_emulated, nvfp4_emulated)?

amd-weisun:
> Thanks for this contribution, I imported it internally and rebased it. The native MXFP4 looks great.
> 
> For the emulated paths (for MXFP4 and NVFP4) I have a few questions:
> 
> * It looks for these paths where we don't have native FP4 support we would leverage dequant the quantized FP4 value into FP8, and leverage FP8 gemm. This would not be equivalent to native FP4 since theres no guarantee the FP8 values could hold the (fp4_quant * scale). Curious if we should document this a bit more, since users could get confused about the different numerical behavior and perhaps leverage a different API for it (e.g. mxfp4_emulated, nvfp4_emulated)?



Hey @cthi , thanks for flagging this out.
The motivation for nvfp4 dequant and fallback on fp8 on AMD is for compatibility.
I would like to hear you opinions about the emulation path, do you think they are necessary? would you prefer to drop the emulation path or put more docs/warnings when emulation paths are used? I am open to both solutions.

jwfromm:
> > Thanks for this contribution, I imported it internally and rebased it. The native MXFP4 looks great.
> > For the emulated paths (for MXFP4 and NVFP4) I have a few questions:
> > 
> > * It looks for these paths where we don't have native FP4 support we would leverage dequant the quantized FP4 value into FP8, and leverage FP8 gemm. This would not be equivalent to native FP4 since theres no guarantee the FP8 values could hold the (fp4_quant * scale). Curious if we should document this a bit more, since users could get confused about the different numerical behavior and perhaps leverage a different API for it (e.g. mxfp4_emulated, nvfp4_emulated)?
> 
> Hey @cthi , thanks for flagging this out. The motivation for nvfp4 dequant and fallback on fp8 on AMD is for compatibility. I would like to hear you opinions about the emulation path, do you think they are necessary? would you prefer to drop the emulation path or put more docs/warnings when emulation paths are used? I am open to both solutions.

I personally think its fine to have it purely as a fallback. If i'm reading Chris' comment right, he's just asking that we note that this is not a numerically equivalent NVFP4 kernel in a docstring.

test/gemm/gemm_test.py
```diff
@@ -2448,13 +2449,122 @@ def test_grouped_gemm_2d_3d(
...
+        torch.testing.assert_close(out_mxfp4, out_bf16, atol=8.0e-2, rtol=8.0e-2)
+
     @parameterized.expand(
         [
             (1, 256, 256, 2048),  # small
             (4, 500, 1024, 2048),  # medium
             (16, 3500, 6144, 3584),  # large
         ]
     )
+    @unittest.skipIf(
```
cthi: Duplicate? Also can use `@skipUnlessRocm`?

test/gemm/gemm_test.py
```diff
@@ -2448,13 +2449,122 @@ def test_grouped_gemm_2d_3d(
 
         torch.testing.assert_close(out_mxfp4, out_bf16, atol=8.0e-2, rtol=8.0e-2)
 
+    @parameterized.expand(
+        itertools.product(
+            [1, 4],  # G
+            [250, 500],  # M
+            [256, 1024],  # N
+            [2048, 3584],  # K
```
cthi: Instead of `itertools.product` please follow the new convention of using some sample cases to avoid combinatorial test case explosion.

test/gemm/gemm_test.py
```diff
@@ -2448,13 +2449,122 @@ def test_grouped_gemm_2d_3d(
...
+
+        out_mxfp4 = torch.ops.mslk.f4f4bf16_grouped_stacked(
+            xq, wq, x_scale, w_scale, M_sizes
+        )
+        self.assertTrue(out_mxfp4.isfinite().all(), "output contains non-finite values")
+        torch.testing.assert_close(out_mxfp4, out_bf16, atol=8.0e-2, rtol=8.0e-2)
+
+    @skipUnlessRocm()
+    @skipUnlessGfxArch("gfx950")
+    def test_grouped_stacked_gemm_uneven(self) -> None:
```
cthi: Instead of having separate test for this, why not fold into above into `test_grouped_stacked_gemm`?

mslk/gemm/triton/f4f4bf16.py
```diff
@@ -0,0 +1,737 @@
...
+        assert offsets.dtype == torch.int32 and offsets.dim() == 1
+
+        if WQ.dim() != 3:
+            raise RuntimeError(
+                "ROCm f4f4bf16_grouped_mm only supports 2D-3D grouped GEMM "
+                "(XQ [total_M, K//2], WQ [G, K//2, N] after caller transpose). "
+                "2D-2D K-grouped layout is CUDA-only."
+            )
+
+        total_M = XQ.shape[0]
```
cthi: why total_m only set when w_scale.dtype != torch.uint8 ?

cthi (commented):
left some comments, thanks

cthi:
@jwfromm yeah, i think adding emulation/alternative recipe support is ok, but I think we should do it in a way that's very clear (otherwise it could get confusing since the recipe would be prone to more errors with outliers)

PR #384, opened 2026-06-11
Fix formatter and linter in OSS to be fbcode compliant

cthi:
Summary: The linter we use in OSS has drifted a bit from internally within Meta, this tries to fix it and make it working again.

Differential Revision: D108302610


bottler (approved):

PR #392, opened 2026-06-17
[ROCm] Add Triton FP8 groupwise (grouped/non-grouped) GEMM kernels

avbokovoy:
## Summary

1. New: f8f8bf16_groupwise. ROCm Triton FP8xFP8 GEMM with groupwise scaling, matching the CUDA CUTLASS schema. Registered automatically for ROCm via mslk/gemm/__init__.py.

2. New: f8f8bf16_groupwise_grouped. ROCm Triton grouped variant with 2D input and 3D weights. CUDA-graph-safe via a  recompute cache keyed on tensor pointer.

3. Fix: CK dtype checks in blockwise, rowwise, rowwise-batched, rowwise-grouped, and rowwise-preshuffle kernels hardcoded float8_e4m3fnuz. Relaxed to XQ.dtype() == WQ.dtype() so gfx950 OCP FP8 (float8_e4m3fn) tensors no longer fail the assertion.

4. Fix: supports_float8_fnuz() previously returned True for all capability >= (9,4), making gfx950 (MI350X) use fnuz which is the slow emulation path. Fixed to return False for (9,5)+. Env var MSLK_ROCM_FORCE_FP8FNUZ_TYPE=1 provided as an escape hatch for debugging and interop.

## Test plan

- [x] Run pytest test/gemm/gemm_test.py -k test_f8f8bf16_groupwise on gfx942 and gfx950.
- [x] Run pytest test/gemm/gemm_test.py -k test_f8f8bf16_groupwise_grouped on gfx942 and gfx950.
- [x] Verify existing FP8 rowwise and blockwise tests pass on gfx950 with OCP tensors.
- [x] Verify MSLK_ROCM_FORCE_FP8FNUZ_TYPE=1 restores the fnuz path on gfx950.

### DIFF
--- a/bench/gemm/gemm_ops.py
+++ b/bench/gemm/gemm_ops.py
@@ -31,6 +31,12 @@ from mslk.quantize.triton.fp8_quantize import (
 )
 from mslk.utils.device import is_cuda, is_gfx942, is_gfx950, is_rocm
 from mslk.utils.triton.fp8_utils import get_fp8_constants
+from mslk.utils.device import (
+    is_cuda,
+    is_gfx942,
+    is_gfx950,
+    is_rocm,
+)
 
 try:
     from tinygemm.utils import group_quantize_tensor
@@ -87,7 +93,7 @@ class Accelerator(Enum):
     NVIDIA_SM90 = auto()
     NVIDIA_SM100 = auto()
     NVIDIA_SM103 = auto()
-    AMD_MI300X = auto()
+    AMD_GFX942 = auto()
     AMD_GFX950 = auto()
 
 
@@ -113,7 +119,7 @@ def get_current_accelerator() -> Accelerator | None:
         if is_gfx950():
             return Accelerator.AMD_GFX950
         if is_gfx942():
-            return Accelerator.AMD_MI300X
+            return Accelerator.AMD_GFX942
     elif is_cuda():
         major, minor = torch.cuda.get_device_capability()
         if major == 9 and minor == 0:
@@ -724,7 +730,7 @@ class FP8Rowwise(GemmOpBase):
             Accelerator.NVIDIA_SM90,
             Accelerator.NVIDIA_SM100,
             Accelerator.NVIDIA_SM103,
-            Accelerator.AMD_MI300X,
+            Accelerator.AMD_GFX942,
         }
 
     @property
@@ -753,7 +759,7 @@ class FP8RowwisePreshuffle(FP8Rowwise):
 
     @property
     def supported_accelerators(self) -> set[Accelerator]:
-        return {Accelerator.AMD_MI300X}
+        return {Accelerator.AMD_GFX942}
 
     @property
     def supported_gemm_types(self) -> set[GemmType]:
@@ -1167,7 +1173,7 @@ class FP8RowwiseGrouped(GemmOpBase):
             Accelerator.NVIDIA_SM90,
             Accelerator.NVIDIA_SM100,
             Accelerator.NVIDIA_SM103,
-            Accelerator.AMD_MI300X,
+            Accelerator.AMD_GFX942,
         }
 
     @property
@@ -1201,7 +1207,7 @@ class FP8RowwiseGrouped2D3D(FP8RowwiseGrouped):
 
     @property
     def supported_accelerators(self) -> set[Accelerator]:
-        return {Accelerator.AMD_MI300X}
+        return {Accelerator.AMD_GFX942}
 
     @property
     def supported_gemm_types(self) -> set[GemmType]:
@@ -1245,7 +1251,7 @@ class TorchFP8RowwiseGrouped(FP8RowwiseGrouped2D3D):
     def supported_accelerators(self) -> set[Accelerator]:
         return {
             Accelerator.NVIDIA_SM90,
-            Accelerator.AMD_MI300X,
+            Accelerator.AMD_GFX942,
         }
 
     @property
@@ -1258,7 +1264,7 @@ class TorchFP8RowwiseGrouped(FP8RowwiseGrouped2D3D):
 
 
 @register_gemm_op
-class CutlassFP8GroupwiseGrouped(GemmOpBase):
+class FP8GroupwiseGrouped(GemmOpBase):
     """
     FP8 grouped matmul with groupwise scaling.
     """
@@ -1293,7 +1299,7 @@ class CutlassFP8GroupwiseGrouped(GemmOpBase):
 
     @property
     def supported_accelerators(self) -> set[Accelerator]:
-        return {Accelerator.NVIDIA_SM90}
+        return {Accelerator.NVIDIA_SM90, Accelerator.AMD_GFX950, Accelerator.AMD_GFX942}
 
     @property
     def supported_gemm_types(self) -> set[GemmType]:
@@ -1337,7 +1343,7 @@ class FP8RowwiseBatched(GemmOpBase):
             Accelerator.NVIDIA_SM90,
             Accelerator.NVIDIA_SM100,
             Accelerator.NVIDIA_SM103,
-            Accelerator.AMD_MI300X,
+            Accelerator.AMD_GFX942,
         }
 
     @property
@@ -1456,7 +1462,7 @@ class FP8Blockwise(GemmOpBase):
     def supported_accelerators(self) -> set[Accelerator]:
         return {
             Accelerator.NVIDIA_SM90,
-            Accelerator.AMD_MI300X,
+            Accelerator.AMD_GFX942,
         }
 
     @property
@@ -1469,7 +1475,7 @@ class FP8Blockwise(GemmOpBase):
 
 
 @register_gemm_op
-class CutlassFP8Groupwise(GemmOpBase):
+class FP8Groupwise(GemmOpBase):
     """
     FP8 matmul with groupwise scaling.
     """
@@ -1496,7 +1502,7 @@ class CutlassFP8Groupwise(GemmOpBase):
 
     @property
     def supported_accelerators(self) -> set[Accelerator]:
-        return {Accelerator.NVIDIA_SM90}
+        return {Accelerator.NVIDIA_SM90, Accelerator.AMD_GFX950, Accelerator.AMD_GFX942}
 
     @property
     def supported_gemm_types(self) -> set[GemmType]:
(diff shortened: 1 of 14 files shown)


bench/gemm/gemm_ops.py
```diff
@@ -109,8 +110,12 @@ def get_current_accelerator() -> Accelerator | None:
         raise Exception("Cannot run gemm_bench without accelerator.")
 
     if torch.version.hip is not None:
-        device_name = torch.cuda.get_device_name()
-        if "MI300X" in device_name.upper():
+        props = torch.cuda.get_device_properties(0)
+        gcn_arch = getattr(props, "gcnArchName", "").upper()
+        device_name = torch.cuda.get_device_name().upper()
+        if "MI350" in device_name or "GFX950" in gcn_arch:
```
cthi: Quick question, so the way we were checking for gfx arch, (that I recently refactor into here: https://github.com/meta-pytorch/MSLK/blob/e6b95a735f5652638762ef9de67f60f5b759205a/mslk/utils/device.py#L52-L82

is a bit different, e.g. we don't check for MI350, we only check for GFX950. Is it the case that some devices would only report MI350 (but not GFX950)?

csrc/gemm/ck/fp8_rowwise_batched/fp8_rowwise_batched_gemm.hip
```diff
@@ -210,9 +210,8 @@ at::Tensor f8f8bf16_rowwise_batched(
       XQ.dim() == 3 && WQ.dim() == 3,
       "Inputs must have 3 dimensions, with the first being batch.");
   TORCH_CHECK(
-      (XQ.dtype() == at::kFloat8_e4m3fnuz) &&
-          (WQ.dtype() == at::kFloat8_e4m3fnuz),
-      "Inputs must be type float8_e4m3fnuz.");
+      XQ.dtype() == WQ.dtype(),
```
cthi, line 213: For these CK kernels, do they work for gfx950 out of the box? I think we currently only compile for gfx942.  But even if we added gfx950 to the arch list, would they work?

bench/gemm/gemm_ops.py
avbokovoy: Not really. Actually, `MI350 in device_name` check is obsolete as `torch.cuda.get_device_name().upper()` will always return `"AMD RADEON GRAPHICS"`. So I just followed the way it's currently implemented by specifying `MI350` check. We should rely on "gfx*" identification instead. For the reference, here's the list of ROCm archs and particular GPUs:
https://rocm.docs.amd.com/en/latest/reference/gpu-arch-specs.html

cthi: Ah that makes sense, thanks for clarifying. Feel free to leave this as is, I will also migrate the gemm_ops.py script to use the shared utility after.

avbokovoy:
Fixed triton compilation error (https://github.com/meta-pytorch/MSLK/pull/392/commits/6f7c222b73e6e922d7dee501e38fbe54834e7f75),  reworked get_current_accelerator() in `bench/gemm/gemm_ops.py` to reuse new `device.py` functions, rebased the branch on current main

avbokovoy:
Seems like triton compilation error wasn't fixed. I'll have closer look

avbokovoy:
All the issues should be solved. The branch is up-to-date with current main branch.

cc: @cthi, @jwfromm

csrc/gemm/ck/fp8_rowwise_batched/fp8_rowwise_batched_gemm.hip
cthi: @avbokovoy  wondering for ^ what your thoughts are? thanks.

mslk/gemm/triton/fp8_groupwise_gemm.py
```diff
@@ -0,0 +1,291 @@
...
+
+from typing import List, Optional
+
+import torch
+import triton  # @manual
+import triton.language as tl  # @manual
+from triton import Config  # @manual
+
+# Scale group size in K: must match the quantize_fp8_group/block group_size.
+_SCALE_GROUP_K: int = 128
```
cthi, line 42: IIRC, the group size is configurable? Shall we pass this as an arg?

mslk/gemm/triton/fp8_groupwise_grouped_gemm.py
```diff
@@ -0,0 +1,322 @@
...
+
+from typing import Dict, List, Optional, Tuple
+
+import torch
+import triton  # @manual
+import triton.language as tl  # @manual
+from triton import Config  # @manual
+
+# Scale group size in K: must match the quantize_fp8_group/block group_size.
+_SCALE_GROUP_K: int = 128
```
cthi, line 40: Should be argument to kernel?

mslk/quantize/triton/fp8_quantize.py
```diff
@@ -903,15 +903,15 @@ def _kernel_quantize_fp8_group(
         # Scale and quantize.
         a_scale = MAX_FP8 / group_max
         scale_chunk_offset = scale_k_offset + k * GROUP_LOAD
-
+        valid = tl.arange(0, GROUP_LOAD) + k * GROUP_LOAD < NUM_GROUPS
```
cthi: nit: maybe keep this as `valid = scale_chunk_offset < NUM_GROUPS` ? Since we already define the `scale_chunk_offset` above

cthi (commented):
Largely looks good to me thanks. I think you just need to rebase.

One thing unresolved thing is the CK C++ kernels, does it actually supports MI350X out of the box?

csrc/gemm/ck/fp8_rowwise_batched/fp8_rowwise_batched_gemm.hip
avbokovoy: > For these CK kernels, do they work for gfx950 out of the box?

@cthi Yes, they work out of the box. CK handles target arch during compilation and uses GPU native formats/hardware operations (unless explicitly manipulated with compile flags, which is mostly not the case for MSLK, but I'll elaborate on that a bit later). The problem with current upstream approach is that it forces both gfx942 and gfx950 to use `fp8fnuz` format, which is native for gfx942 and not for gfx950 (it uses OCP). 

So, on gfx950 forcing fnuz format will silently break the numerical correctness. It most likely won't be caught with unit-tests (as it is now, maybe with some flaky atol/rtol threshold failures, exceeding it by marginal value), but it will (in fact it already does) yield significant numerical mismatches on large cases (any close to real prefill/decode-like shapes) and in some of the benchmarks (unusual sim/SQNR values). The problem may also occur when the computation is done around fp8 max/NaN/+-zero values.

With all that being said, we tried to address the mentioned above issue in this PR.

> I think we currently only compile for gfx942

That one is only partially true. ROCm FMHA as a separate target library is compiled for gfx942 with it's own compile flags (it's hardcoded in CMake file). The rest follows build target arch specified in `AMDGPU_TARGETS` variable. For ROCm CI pipeline the arches are `gfx908,gfx90a,gfx942,gfx950` and that's probably one of the reasons why builds take ages. And there's also an ambiguity with `PYTORCH_ROCM_ARCH`. But anyway, seems like FMHA has nothing to do with fp8, so it's more a performance consideration.

>  But even if we added gfx950 to the arch list, would they work?

They already work :)

But there's also one factor that might decrease importance of some of the mentioned above problems with CK kernels. I'll share more details in another communication channel

mslk/quantize/triton/fp8_quantize.py
avbokovoy: You are right, done in https://github.com/meta-pytorch/MSLK/pull/392/commits/b8f9bf1c2d1b79e0fed50ea1fb9394237978ebcf

mslk/gemm/triton/fp8_groupwise_gemm.py
avbokovoy: The input layout dictates the `_SCALE_GROUP_K` value and it matches current CUTLASS implementation. Not sure that making it as a kernel argument is necessary, since it's not exposed to the public API. But if you want to avoid using global private variables in python code, just let me know, I'll fix it

mslk/gemm/triton/fp8_groupwise_grouped_gemm.py
avbokovoy: https://github.com/meta-pytorch/MSLK/pull/392#discussion_r3497672113

avbokovoy:
> Largely looks good to me thanks. I think you just need to rebase.
> 
> One thing unresolved thing is the CK C++ kernels, does it actually supports MI350X out of the box?

@cthi Thanks for your review. I addressed your comments. More or less detailed answer for your question is in https://github.com/meta-pytorch/MSLK/pull/392#discussion_r3497648789. Let me know if you have any further questions

csrc/gemm/ck/fp8_rowwise_batched/fp8_rowwise_batched_gemm.hip
cthi: Ah I see, thanks for explaining this!

mslk/gemm/triton/fp8_groupwise_gemm.py
cthi: Ah right, I forgot this is fixed currently. Its fine as is.

mslk/gemm/triton/fp8_groupwise_grouped_gemm.py
```diff
@@ -0,0 +1,322 @@
...
+    if output is None:
+        output = torch.empty((TotalM, N), dtype=torch.bfloat16, device=XQ.device)
+
+    if TotalM == 0 or N == 0 or K == 0 or G == 0:
+        return output
+
+    # Precompute M_starts (cumulative row offsets) and max_M on CPU.
+    # Results are cached so CUDA graph capture (which forbids device-to-host
+    # transfers) can reuse the values computed on the first non-capturing call.
+    cache_key = (M_sizes.data_ptr(), G, N, K)
```
cthi: @avbokovoy so, I don't think this is valid PyTorch code. Cuda caching allocator could re-use the data pointer. Is there some reason we cannot just have another kernel to write the metadata you need? Or compute this metadata on GPU? Cudagraph capture should properly capture those?

avbokovoy: @cthi You are right. Fixed it in [ea7c926](https://github.com/meta-pytorch/MSLK/pull/392/commits/ea7c926e192ed3b9de818bee81ae5f84953ee2fa). Also rebased the branch

PR #393, opened 2026-06-17
Add Autograd Backward Support for FP4 GEMM Custom Ops

colalb1:
Adds backward pass support for the three FP4 GEMM custom ops (`mslk::f4f4bf16`, `mslk::f4f4bf16_grouped_mm`, `mslk::f4f4bf16_ultra_grouped_mm`) so that `.backward()` works through them during FP4 training. Previously these operations were forward-only.

**Implementation Method:** 
Dequantize-then-compute in BF16. FP4 (E2M1, only 8 representable magnitudes) cannot carry gradient signal, so the backward dequantizes saved packed FP4 inputs to BF16 and computes standard matmul gradients:
- `grad_X = grad_output @ W_bf16`
- `grad_W = grad_output.T @ X_bf16`

This matches the FP8‑forward / BF16‑backward pattern in the repository.

## Files changed

- `mslk/gemm/fp4_autograd.py` (new): Dequantization helpers + `register_autograd` for all 3 ops
- `mslk/gemm/__init__.py`: Import `fp4_autograd` to trigger registration
- `test/gemm/gemm_test.py`: `MXFP4BackwardTests` and `NVFP4BackwardTests` classes


colalb1:
I put the FP4 autograd logic in its own file rather than inlining into `_meta.py`. The FMHA precedent keeps autograd in the same file as the forward. The forward is in C++ and `_meta.py` is just a shape stub covering GEMM operations.

I put all of the dequantization and gradient logic in the same file, but this felt a bit awkward. Please let me know if I should organize this differently.

Thanks!

jwfromm (approved):
Overall looks great, thank you for the contribution! I think youll just need to rebase and do a little bit of merge cleanup and we can land.

colalb1:
> Overall looks great, thank you for the contribution! I think youll just need to rebase and do a little bit of merge cleanup and we can land.

@jwfromm Thanks! I made the necessary changes. If there is anything else, just lmk.

q10:
@colalb1 could you rebase this PR onto latest main?

PR #396, opened 2026-06-18
Enable Pyrefly for fbcode/mslk

maggiemoss:
Summary:
X-link: https://github.com/facebookexternal/MSLK-NV/pull/346

Migrates fbcode/mslk from Pyre to Pyrefly by removing the root set_pyrefly(False) opt-out so the tree inherits set_pyrefly(True) from the fbcode root default. Suppressed 106 pre-existing type errors that Pyrefly surfaces but Pyre did not, with annotation-only # pyrefly: ignore comments; no runtime changes.

Differential Revision: D108818855


grievejia (approved):

PR #407, opened 2026-06-26
[ROCm] Fix MX8MX4 ROCm Triton kernel: use plain row-major w_scale layout after triton_quantize_mx4_unpack change

avbokovoy:
## Summary

Commit 14526d1 (Add Triton ROCm backend for MXFP4 GEMM ops MI350X) changed triton_quantize_mx4_unpack on ROCm to return w_scale in a plain [N, K//32] row-major layout rather than the padded and swizzled TCGen5 flat buffer that NVIDIA uses. The mx8mx4bf16 and mx8mx4bf16_grouped_mm Triton kernels in mx8mx4_gemm.py were not updated alongside that change and continued to read WS_ptr using the _to_blocked index arithmetic, which maps to wrong byte offsets in the new plain layout. The result was incorrect scale values applied during the GEMM, causing numerical failures across all MX8MX4 test shapes on ROCm.

## Changes

Both _mx8mx4bf16_kernel and _mx8mx4bf16_grouped_kernel now load w_scale via stride-based row-major indexing. The dense kernel uses n2d * stride_wsn + ks2d * stride_wsk and the grouped kernel uses g * stride_wsg + n2d * stride_wsn + ks2d * stride_wsk. Each kernel gains two new stride parameters for WS_ptr. The Python wrappers reshape w_scale to [N, K//32] (dense) or [G, N, K//32] (grouped) before the kernel call and pass the resulting strides. The x_scale path is untouched because tests and benchmarks explicitly apply _to_blocked() to the MX8 activation scales before calling the op.

## Scope and non-impact

The mx8mx4* kernel path is ROCm-only (guarded by torch.version.hip is not None). No NVIDIA code is touched. All f4f4bf16* ops, the fp4_quantize_test, and every other caller of triton_quantize_mx4_unpack route through different kernels that are unaffected. The MX8MX4Groupwise and MX8MX4GroupwiseGrouped2D3D benchmarks call the same ops through torch.ops.mslk and receive the corrected output with no interface change.

## Test plan

Ran the previously failing tests on ROCm (gfx950): MX8MX4Tests::test_gemm_2, test_mx8mx4_grouped_gemm_2d_3d_0 through _3, and test_mx8mx4_grouped_gemm_2d_3d_empty_groups. All pass. No other test classes are affected by this change.

### DIFF
--- a/mslk/gemm/triton/mx8mx4_gemm.py
+++ b/mslk/gemm/triton/mx8mx4_gemm.py
@@ -127,7 +127,7 @@ def _mx8mx4bf16_kernel(
     XQ_ptr,  # [M, K]     float8_e4m3  (raw bytes; reinterpreted in kernel)
     WQ_ptr,  # [N, K//2]  uint8        (two MXFP4 nibbles per byte)
     XS_ptr,  # flat uint8  _to_blocked layout (E8M0 exponents for XQ)
-    WS_ptr,  # flat uint8  _to_blocked layout (E8M0 exponents for WQ)
+    WS_ptr,  # [N, K//32] uint8  plain row-major (E8M0 exponents for WQ)
     Out_ptr,  # [M, N]     bfloat16
     M,
     N,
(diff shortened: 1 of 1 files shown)


jwfromm (approved):
LGTM, thank you for the fix!

PR #408, opened 2026-06-26
Add FlyDSL JIT backend support for ROCm

aryaman-gupta:
## Summary
Enables JIT compilation of FlyDSL kernels in MSLK as an optional ROCm backend, so FlyDSL kernels can be exposed through MSLK ops.

- Generic support module (`mslk/utils/flydsl.py`): availability detection (guarded on FlyDSL's supported archs via `SMEM_CAPACITY_MAP`), an actionable `require_flydsl()`, and a compile-and-cache dispatch shim.
- `flydsl` `extras_require` entry; `install_requires` stays numpy-only per existing policy.
- `mslk/gemm/flydsl/` kernel-package scaffold and ROCm CI install wiring (`utils_flydsl.bash`) mirroring the existing Triton setup.

Infrastructure only — no kernels here. Kernel modules register inline via `@torch.library.impl` guarded by `is_flydsl_available()`, matching the Triton backend.

## Installation

FlyDSL ships on public PyPI; MSLK ships from the PyTorch index. Use `--extra-index-url` (not `--index-url`) so pip searches both.

### Via pip (ROCm)
```bash
# Works today: install MSLK + FlyDSL explicitly
pip install mslk "flydsl==0.2.2" --extra-index-url https://download.pytorch.org/whl/rocm7.1/

# Once this change ships in a released MSLK wheel, the extra pulls FlyDSL in:
pip install "mslk[flydsl]" --extra-index-url https://download.pytorch.org/whl/rocm7.1/
```

### From source
Uses MSLK's standard from-source flow (`ci/integration/mslk_oss_build.bash` creates a conda env and installs the matched build deps). `BUILD_VARIANT=rocm` is required — it defaults to `cuda`, and FlyDSL is only installed on the ROCm variant.
```bash
git clone https://github.com/meta-pytorch/MSLK && cd MSLK
git submodule update --init --recursive

# One-time: build the ROCm conda env and install all deps incl. FlyDSL (slow).
# BUILD_ROCM_VERSION is optional (defaults to 7.0); set it to match your ROCm.
BUILD_VARIANT=rocm ./ci/integration/mslk_oss_build.bash

# After setup, activate the env and rebuild MSLK on each change. The exact
# env name is printed by the script, e.g.:
conda activate build-py3.14-torchnightly-rocm7.0
python setup.py install
```

**Python-only variant** (of the from-source flow) — skips the C++/CUDA build when you only need the Python/FlyDSL path:
```bash
MSLK_PYTHON_ONLY=1 pip install -e ".[flydsl]"
```

## Verify the backend
```python
from mslk.utils.flydsl import is_flydsl_available
print(is_flydsl_available())   # True on a supported ROCm GPU with FlyDSL installed
```

## Test plan
- [x] `pytest test/utils/flydsl_test.py` — availability detection, require raise/pass, dispatch compile-then-cache (mocked; runs without GPU/FlyDSL). 5/5 pass.
- [x] Verified on gfx950 (ROCm 7.2, torch 2.10): `flydsl==0.2.2` from public PyPI imports and JIT-compiles.
- [ ] ROCm CI installs `flydsl==0.2.2` and imports it (smoke test in `utils_flydsl.bash`) — pending a CI run on these branches.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -694,6 +694,7 @@ def main(argv: List[str]) -> None:
     # where cu126 changes to match the cuda version..
     extras_require = {
         "flash3": ["flash-attn-3"],
+        "flydsl": ["flydsl==0.2.2"],
     }
 
     packages = setuptools.find_packages()
--- a/setup.py
+++ b/setup.py
@@ -694,8 +694,12 @@ def main(argv: List[str]) -> None:
     # where cu126 changes to match the cuda version..
     extras_require = {
         "flash3": ["flash-attn-3"],
-        "flydsl": ["flydsl==0.2.2"],
     }
+
+    # FlyDSL is a ROCm-only backend; also offered on variant-agnostic
+    # python-only builds.
+    if _PYTHON_ONLY or build.variant() == "rocm":
+        extras_require["flydsl"] = ["flydsl==0.2.2"]
 
     packages = setuptools.find_packages()
 
--- a/.github/workflows/_mslk_rocm_test.yml
+++ b/.github/workflows/_mslk_rocm_test.yml
@@ -137,6 +137,9 @@ jobs:
     - name: Install MSLK dependencies (pip)
       run: . $PRELUDE; install_mslk_deps $BUILD_ENV
 
+    - name: Install FlyDSL
+      run: . $PRELUDE; install_flydsl_pip $BUILD_ENV
+
     - name: Download Wheel Artifact from GHA
       if: ${{ inputs.mslk-channel-version == '' }}
       uses: actions/download-artifact@v4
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -700,6 +694,7 @@ def get_tag(self):
     # where cu126 changes to match the cuda version..
     extras_require = {
         "flash3": ["flash-attn-3"],
+        "flydsl": ["flydsl==0.2.2"],
```
q10: Should this inclusion be guarded for the case when we're building non-ROCm MSLK wheels?

aryaman-gupta: Sure, done in https://github.com/meta-pytorch/MSLK/pull/408/commits/170f31a78c46c9aa3bf8f408553cb59f0c59d019

Included the Python-only builds since they don't detect ROCm/CUDA build variant. Since this is an extra, it'll anyway only be installed if you specifically request for it

PR #412, opened 2026-06-30
[ROCm] Fix get_clock_rate_in_khz() and _fp8_clamp() functions

avbokovoy:
Current ROCm CI fails with missing `pynvml` dependency and mismatched fp8 format on ROCm. This PR addresses those issues


avbokovoy:
@cthi @jwfromm Ok, seems like there are some new problems poped up. Let me address them in this PR

q10:
@avbokovoy we also happened to open up a similar PR, https://github.com/meta-pytorch/MSLK/pull/413 and its followup https://github.com/meta-pytorch/MSLK/pull/419 , which, when combined, seems to have resolved the ROCm build and test issues

PR #431, opened 2026-07-09
Add FlyDSL AOT pre-compilation support for ROCm

aryaman-gupta:
## Summary
Adds AOT pre-compilation for FlyDSL kernels, building on the FlyDSL JIT backend (#408). Pre-compiles selected kernels in selected configurations into FlyDSL's on-disk cache so the front-end compile is skipped at runtime; cache misses fall back to JIT.

- `mslk/utils/flydsl_aot.py`: config collection + parallel compile harness, plus a `python -m mslk.utils.flydsl_aot` entrypoint. AOT-eligible kernel modules declare `AOT_CONFIGS`, `AOT_ARCHS`, and `compile_aot_config(config, arch)`.
- `mslk/utils/flydsl.py`: `configure_runtime_cache()` points FlyDSL at the bundled cache on import; `MSLK_FLYDSL_RUN_ONLY=1` forbids the JIT fallback (e.g. for CUDA graph capture).
- ROCm build workflow runs the harness (under `MSLK_PYTHON_ONLY=1`) right before wheel packaging so the cache is bundled via `package_data`. The cache is gitignored and never committed.

The worker pool uses the `spawn` start method: workers compile FlyDSL kernels which initialize CUDA, and forking a parent that has already initialized CUDA fails re-initialization.

**Design notes.** This follows MSLK's existing offline-codegen convention (like the CK `kernel_gen.py` flow): generation is an offline/CI step, and the build just packages the artifact. The FlyDSL cache stores lowered IR, not a standalone binary, so FlyDSL remains a runtime dependency — same as the JIT path.

Infrastructure only: no kernel declares `AOT_CONFIGS` yet, so the harness is a no-op until a kernel opts in.

## Test plan
- [x] `pytest test/utils/flydsl_aot_test.py` — job collection, COMPILE_ONLY/cache-dir env set during compile, env restored after. Passes.
- [x] `MSLK_PYTHON_ONLY=1 python -m mslk.utils.flydsl_aot` runs cleanly (no-op with zero registered kernels).
- [x] Verified end-to-end on gfx950 (ROCm 7.2, torch 2.10) with a sample kernel: AOT compile populates the cache, the op is served from cache, and `MSLK_FLYDSL_RUN_ONLY=1` errors on a non-precompiled shape (proving cache-hit-no-JIT).
- [ ] ROCm CI build step bundles a populated cache into the wheel — pending a live CI run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ mslk/version.py
 **/__pycache__
 **/*.so
 *.profraw
-mslk/utils/flydsl_aot_cache/
+mslk/flydsl/aot_artifacts/
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ mslk/version.py
 **/__pycache__
 **/*.so
 *.profraw
+mslk/utils/flydsl_aot_cache/
--- a/mslk/utils/flydsl.py
+++ b/mslk/utils/flydsl.py
@@ -17,6 +17,7 @@ and op-specific logic stays in the kernel modules.
 
 import functools
 import importlib.util
+import os
 from typing import Any, Callable
 
 _INSTALL_HINT: str = (
@@ -50,6 +51,27 @@ def require_flydsl() -> None:
         raise RuntimeError(_INSTALL_HINT)
 
 
+# Bundled AOT cache shipped inside the package (populated at build time).
+# Shared across all FlyDSL kernel categories (gemm, moe, ...), so it lives
+# alongside the FlyDSL support module rather than under any kernel domain.
+_BUNDLED_AOT_CACHE: str = os.path.join(os.path.dirname(__file__), "flydsl_aot_cache")
+
+
+def configure_runtime_cache() -> None:
+    """Point FlyDSL at the bundled AOT cache, if present and not overridden.
+
+    Cache hits skip the front-end compile; misses fall back to JIT. Set
+    ``MSLK_FLYDSL_RUN_ONLY=1`` to forbid the JIT fallback (e.g. for CUDA
+    graph capture), which makes FlyDSL raise on a cache miss instead.
+    """
+    if "FLYDSL_RUNTIME_CACHE_DIR" not in os.environ and os.path.isdir(
+        _BUNDLED_AOT_CACHE
+    ):
+        os.environ["FLYDSL_RUNTIME_CACHE_DIR"] = _BUNDLED_AOT_CACHE
+    if os.environ.get("MSLK_FLYDSL_RUN_ONLY", "0") == "1":
+        os.environ["FLYDSL_RUNTIME_RUN_ONLY"] = "1"
+
+
 def run_compiled(launcher: Callable[..., Any], *args: Any) -> None:
     """Dispatch a FlyDSL ``@flyc.jit`` host launcher with ``args``.
 
@@ -63,3 +85,6 @@ def run_compiled(launcher: Callable[..., Any], *args: Any) -> None:
         launcher._mslk_cf = flyc.compile(launcher, *args)  # pyre-ignore[16]
     else:
         cf(*args)
+
+
+configure_runtime_cache()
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -8,3 +8,4 @@ mslk/version.py
 **/__pycache__
 **/*.so
 *.profraw
+mslk/utils/flydsl_aot_cache/
```
cthi: I think it's a bit weird to have the generated artifacts here, perhaps `mslk/flydsl/aot_artifacts` ?

mslk/utils/flydsl.py
```diff
@@ -50,6 +51,27 @@ def require_flydsl() -> None:
...
+# Shared across all FlyDSL kernel categories (gemm, moe, ...), so it lives
+# alongside the FlyDSL support module rather than under any kernel domain.
+_BUNDLED_AOT_CACHE: str = os.path.join(os.path.dirname(__file__), "flydsl_aot_cache")
+
+
+def configure_runtime_cache() -> None:
+    """Point FlyDSL at the bundled AOT cache, if present and not overridden.
+
+    Cache hits skip the front-end compile; misses fall back to JIT. Set
+    ``MSLK_FLYDSL_RUN_ONLY=1`` to forbid the JIT fallback (e.g. for CUDA
```
cthi: I think we could call this `MSLK_FLYDSL_DISABLE_JIT` to make it more clear ?

cthi (commented):
Few questions about FlyDSL semantics of the JIT cache:

- Is the JIT cache artifacts deterministic? E.g. given same program would FlyDSL always produce exact same artifacts?
- How long does AOT compile take roughly per kernel instance? E.g. an instance of an MNK gemm.

.gitignore
aryaman-gupta: Good point — done in 1800db4. The generated artifacts now live under a dedicated `mslk.flydsl` package at `mslk/flydsl/aot_artifacts/`, rather than alongside the support module in `mslk/utils/`.

Since we now have a `mslk.flydsl` package, would it make sense to also move the FlyDSL support modules there — i.e. `mslk/utils/flydsl.py` → `mslk/flydsl/` and `mslk/utils/flydsl_aot.py` → `mslk/flydsl/`? That would keep all FlyDSL code under one package. I left them in `mslk/utils/` for now to keep this PR focused, but happy to move them if you'd prefer.

mslk/utils/flydsl.py
aryaman-gupta: Renamed to `MSLK_FLYDSL_DISABLE_JIT` in dc2548f.

.gitignore
cthi: Yeah I think we could move them into there after (feel free to do it in a later PR)

aryaman-gupta:
Hi @cthi ,

> Few questions about FlyDSL semantics of the JIT cache:
> 
> * Is the JIT cache artifacts deterministic? E.g. given same program would FlyDSL always produce exact same artifacts?

Determinism holds where it matters for cache reuse: FlyDSL's cache key is a SHA-256 over its version + compiler/native-lib hashes + target arch + kernel source/params. Same version+arch → byte-identical artifacts; a different FlyDSL version or arch
changes the key, so a stale artifact is invalidated and recompiled rather than silently reused. Practical implication: the bundled cache is valid only for the exact FlyDSL build it was compiled against, so we regenerate it at wheel-build time against the pinned flydsl==0.2.2 (CI does this before packaging); on any mismatch it cleanly falls back to JIT.

> * How long does AOT compile take roughly per kernel instance? E.g. an instance of an MNK gemm.

Measured a few representative GEMM/MoE kernels on MI350 — front-end compile is on the order of ~1s per kernel instance (cold). Total AOT time scales with the number of instances, but the harness compiles them in parallel, so wall-clock stays modest.

.gitignore
aryaman-gupta: Sure. We are planning another PR soon, modifying how the FlyDSL dependency is declared for ROCm builds. We can add this change there. Thanks!

PR #436, opened 2026-07-14
Migrate contained C10_LIKELY/C10_UNLIKELY to [[likely]]/[[unlikely]] (mslk)

r-barnes:
Summary:
Migrates fully-contained uses of the `C10_LIKELY`/`C10_UNLIKELY` macros -- where the macro is the entire controlling condition of an `if`/`else if`/`while`/`for` -- to the standard C++20 `[[likely]]`/`[[unlikely]]` branch-prediction attributes, e.g. `if (C10_LIKELY(x))` becomes `if (x) [[likely]]`.

This diff is one of several split out from D111576576 by GitHub project, and contains only the changes under `fbcode/mslk/`. Conditions where the macro is only one clause of a compound condition (`if (a && C10_LIKELY(b))`), `return C10_LIKELY(x);` forms, macro definitions, and `generated` files are deliberately left untouched.

The migration was performed by the `c10_likely_to_attr` codemod, landed as a companion diff.

This diff was authored with the assistance of an AI coding agent.

Differential Revision: D111945036


bottler (approved):

PR #442, opened 2026-07-21
setup.py simplification

cthi:
Summary: as titled

Differential Revision: D113023122


bottler (approved):

PR #443, opened 2026-07-21
Enable Blackwell Ultra Tiles in f4f4bf16

wshwang:
Summary:
Enables the GB300 (Blackwell Ultra, SM103) FP4 path for the non-grouped block-scaled FP4→BF16 GEMM `torch.ops.mslk.f4f4bf16`, covering all three FP4 formats (NVFP4, MXFP4, MXFP4_16). The op keeps a single public entry point with an unchanged signature, so callers are unaffected: on `sm >= 103` (CUDA 13+) it transparently dispatches to a dedicated SM103 "ultra" kernel at runtime, and on B200/SM100 or pre-CUDA-13 builds it runs exactly the existing Sm100 path.

## Arch-aware kernel template (`f4f4bf16/f4f4bf16_common.cuh`)

`_f4f4bf16<...>` gains a `bool UseUltra` template parameter.

- When `UseUltra` and `CUDA_VERSION >= 13000`, the kernel builds for `cutlass::arch::Sm103` with block-scaled ultra schedules named explicitly per format (NVFP4 and MXFP4_16 use `Vs16`, MXFP4 uses `Vs32`), passes `cute::tuple<data, scale>` element pairs to the SM103 `CollectiveBuilder`, and uses `TileShape_K = 768` (hard-required by the SM103 ultra builder). 2SM vs 1SM is derived from the tile/cluster (`TB_M == 256 && cluster_M` even → 2SM, else 1SM).
- When `!UseUltra` the kernel is byte-for-byte the original Sm100 kernel.
- A CUDA-13 gate with an Sm100/`Auto` fallback alias keeps `UseUltra=true` instances compilable on CUDA 12.8 (they are never dispatched there).

## Transparent, format-aware runtime dispatch (`f4f4bf16.cu`)

- `get_device_sm_version(device_index)` — `major*10 + minor` for the given device (100 = B200, 103 = B300), looked up per call on `XQ`'s device.
- `ultra_path_disabled()` — reads `MSLK_F4F4BF16_DISABLE_ULTRA` once; forces the Sm100 path on B300 as a rollback / A-B escape hatch.
- `f4f4bf16(...)` routes to the format-aware SM103 selector when `sm >= 103` (CUDA 13+) and ultra is not disabled; otherwise to the existing Sm100 `get_kernel_via_heuristics`.
- `get_ultra_kernel_via_heuristics(M, N, K, global_scale, mxfp4_block_size)` dispatches by format — NVFP4 (`global_scale` set) / MXFP4_16 (`mxfp4_block_size == 16`) / MXFP4 (block == 32) — to one of three per-format trees (`get_ultra_kernel_via_heuristics_nvfp4` / `_mxfp4_16` / `_mxfp4`), each generated from that format's own autotune sweep so every format picks its own optimum from the 52-tile pool. This format-aware split removes the mid-M MXFP4/MXFP4_16 regressions (per-shape floor ≈0.83) that a single NVFP4-derived tree caused when applied to all formats.
- `get_f4f4bf16_ultra_kernels()` — autotune map of the SM103 ultra (K=768) tiles.
- `get_f4f4bf16_sm103_kernels()` — combined SM103 autotune pool = the Sm100 (K=256) tiles ∪ the ultra (K=768) tiles, so `MSLK_AUTOTUNE_ENABLE` can select per shape from both arch paths. Autotune keeps a separate tuning cache file per FP4 format (format encoded in the filename, keys stay `M_N_K`), so per-format selection matches the heuristic dispatch — different formats can pick different tiles at the same shape. The Sm100 tiles remain the only option for shapes where the wide K=768 ultra tile is inefficient (ultra cannot use a smaller K).

## Tiles (`f4f4bf16/f4f4bf16_ultra_*_768_*.cu`, `f4f4bf16_manifest.cuh`)

39 SM103 ultra tile instances (K=768) spanning the 128-M (1SM) and 256-M (2SM) geometry families with `TB_N` in {128, 192, 256}, all guarded on `CUDA_VERSION >= 13000` and declared in the manifest. All 39 are verified to compile and pass `can_implement` on GB300 for all three formats. The combined SM103 candidate pool is 52 (13 Sm100 + 39 ultra); the three shipped trees reference 26 of them.

## Rollback

`MSLK_F4F4BF16_DISABLE_ULTRA=1` forces the Sm100 path on B300 (A/B escape hatch / rollback).

Reviewed By: cthi

Differential Revision: D112152886


cthi:
closed as merged

PR #447, opened 2026-07-23
Make FlyDSL a required dependency on ROCm builds

aryaman-gupta:
## Summary
Makes FlyDSL a mandatory dependency on ROCm builds (native and python-only, including source installs), consolidates the FlyDSL support code under the `mslk.flydsl` package, and single-sources the version pin.

- **Mandatory on ROCm.** FlyDSL is injected into `install_requires` when the build variant resolves to `rocm`, instead of being an opt-in extra. Native builds use `build.variant()`; python-only builds (including `MSLK_PYTHON_ONLY=1 pip install -e .` source installs) use `_detect_build_variant()`, which now prefers `CU_VERSION` and falls back to a lazy `torch` import — so it resolves the variant without torch imported at module load and degrades to `cpu` when torch is absent. cpu/cuda builds are unchanged (numpy only).
- **`__variant__` stamp untouched.** The python-only wheel still stamps `"python_only"`, so runtime native-library loading (the `.so` skip) is unaffected.
- **Single source of truth for the pin.** The version moves to `ci/flydsl_version.txt`, read by both `setup.py` and `ci/scripts/utils_flydsl.bash`, and is bumped `0.2.2 -> 0.2.4`.
- **Package consolidation (follow-up from #431).** As agreed in [this review thread on #431](https://github.com/meta-pytorch/MSLK/pull/431#discussion_r3559079742), `mslk/utils/flydsl.py` and `flydsl_aot.py` move under `mslk/flydsl/`, split into `common.py` (availability + shared cache path), `jit.py` (runtime dispatch), and `aot.py` (AOT pre-compilation). Tests mirror the split under `test/flydsl/`.
- **Cleanup.** Removes the now-redundant `flydsl` extra and updates the install hint to `pip install flydsl`.
- **Docs & CI install fix.** Updates the README ROCm install commands and adds ROCm build-from-source instructions; also teaches `ci/scripts/utils_pip.bash` to add PyPI as an extra index for ROCm installs/downloads so CI can resolve the new FlyDSL dependency.

> [!IMPORTANT]
> **This changes ROCm install behavior.** Because the ROCm wheel now requires `flydsl` (published on PyPI, not on the PyTorch index), the current README command will fail to resolve the dependency:
> ```bash
> # No longer sufficient — --index-url replaces PyPI, so flydsl can't be found:
> pip install mslk --index-url https://download.pytorch.org/whl/rocm7.1/
> ```
> Users must now also make PyPI available, e.g.:
> ```bash
> pip install mslk \
>   --index-url https://download.pytorch.org/whl/rocm7.1/ \
>   --extra-index-url https://pypi.org/simple
> ```
> The README is updated accordingly.

## Test plan
Verified end-to-end in a `rocm/pytorch:latest` container (torch 2.10 + ROCm 7.2, gfx942):
- [x] `test/flydsl/` unit suite — 7 passed
- [x] Native ROCm metadata includes `Requires-Dist: flydsl==0.2.4` (mandatory, no extra marker); native cpu excludes it
- [x] Python-only ROCm source install (torch.hip set, no `CU_VERSION`) requires `flydsl==0.2.4`
- [x] Python-only `CU_VERSION=rocm6300` requires flydsl; `CU_VERSION=cu126` does not
- [x] `mslk.flydsl.{common,jit,aot}` import and integrate against real flydsl 0.2.4 (availability detection, cache config, AOT no-op harness)
- [x] Update README ROCm install commands + build instructions, and add the PyPI extra-index fallback in CI (`utils_pip.bash`)


ci/scripts/utils_pip.bash
```diff
@@ -227,12 +235,17 @@ install_from_pytorch_pip () {
...
 
   echo "[INSTALL] Attempting to install [${package_name}, ${package_version:-LATEST}] from PyTorch PIP using channel ${pip_channel} ..."
+  if [ -n "${pip_extra_index_args}" ]; then
+    echo "[INSTALL] Adding PyPI fallback for ROCm dependency resolution: ${pip_extra_index_args}"
+  fi
   # shellcheck disable=SC2086
   # NOTE: `python -m pip` (not bare `pip`) ensures we install into the env's own
   # site-packages even if its `pip` resolution falls back to base conda's pip.
-  (exec_with_retries 3 conda run ${env_prefix} python -m pip install ${pip_package} --index-url ${pip_channel}) || return 1
+  (exec_with_retries 3 conda run ${env_prefix} python -m pip install ${pip_package} --index-url ${pip_channel} ${pip_extra_index_args}) || return 1
```
q10: I think if a package happens to be available in both index-url and extra-index-url (such as torch), then pip appears to pull both, and pick the highest version of the two, which might end up confusing things.

Maybe it's better to split installation of flydsl from installation of mslk?

cthi: Yeah I agree with Benson. Can we isolate the flydsl dep to only that pip index?

PR #444, opened 2026-07-22
feat: add FlyDSL batched preshuffle GEMM for FP8 rowwise scaling (WP-G2)

kudomcho:
## Motivation

Add batched FP8 preshuffle GEMM via FlyDSL for gfx950 (MI350). Derivative of WP-G1 — builds on the non-batched preshuffle kernel from PR #434.

Depends on WP-G1 PR #434. Rebased on main after PR #447 merge.

## Technical Details

**New API: `mslk.gemm.flydsl.preshuffle_gemm`**
- `flydsl_preshuffle_batched_gemm(XQ, WQ, x_scale, w_scale, ...)` — batched GEMM with Grid-Z batching (single kernel launch for all B batches)
- Preshuffled weights cached by `data_ptr` — preshuffle once, reuse on subsequent calls
- Registered as `torch.library.impl("mslk::f8f8bf16_rowwise_batched", "CUDA")` on gfx950

**Performance optimizations:**

1. **Grid-Z batching** — uses `gpu.block_id("z")` to index into the batch dimension, launching all B batches in a single kernel with `grid=(gx, gy, B)`. Each batch's buffer resource addresses are offset by `bz * batch_stride_bytes`. Eliminates Python dispatch overhead entirely (was ~258 us for 16 `run_compiled` calls).

2. **XCD swizzle + waves_per_eu tuning** — full parameter sweep across tile configs × `xcd_swizzle` (0,1,2,4) × `waves_per_eu` (0,1,2) identified optimal settings per config. `xcd_swizzle=1` improves L2 cache reuse across chiplets; `waves_per_eu=2` improves scheduling on large shapes.

3. **Profile-guided shape overrides** — `_SHAPE_OVERRIDES_GFX950` lookup table maps `(m_range, N, K)` to sweep-optimal configs for N=1280 and N=8192 shapes where the heuristic picks suboptimal tiles.

4. **Batch-aware occupancy heuristic** — `select_default_config(batch=B)` factors Grid-Z parallelism into the occupancy threshold (`m_tiles * n_tiles * B >= 64`), enabling larger tile configs that were previously rejected.

**Benchmark: `FP8RowwiseBatchedPreshuffleFlyDSL`** — new benchmark class in `bench/gemm/gemm_ops.py` targeting `AMD_GFX950`, gated on `is_flydsl_available()`.

## Test Plan

```bash
# Correctness test
pytest test/gemm/gemm_test.py -k FlyDSLPreshuffleBatchedGemmTest
```

[...]

PR #434, opened 2026-07-10
feat: add FlyDSL preshuffle GEMM for FP8 rowwise scaling (gfx950)

kudomcho:
## Motivation

Add FlyDSL JIT backend for FP8 rowwise preshuffle GEMM on gfx950 (MI350). MSLK's existing `f8f8bf16_rowwise_preshuffle` op only has a CK backend targeting gfx942. FlyDSL provides a Python-only JIT-compiled alternative with faster GPU kernels on most shapes, though Python dispatch overhead reduces net gains on small-M shapes.

## Technical Details

Port the FlyDSL preshuffle GEMM kernel from ROCm/aiter into MSLK as a self-contained module with no aiter dependency.

**New module: `mslk.gemm.flydsl`**
- `flydsl_preshuffle(src)` — weight shuffle into FlyDSL layout (different from `ck_preshuffle`)
- `flydsl_preshuffle_gemm(XQ, WQ, x_scale, w_scale, ...)` — JIT-compile and run the kernel
- `_configs.py` — default tile configs for gfx950 heuristic selection
- `_kernels/` — ported FlyDSL kernel compiler (preshuffle_gemm, mfma_epilogues, mfma_preshuffle_pipeline, tensor_shim)

**Op wiring (gfx950, gated on `is_flydsl_available()`):**
- `torch.ops.mslk.f8f8bf16_rowwise` → FlyDSL (bf16 output)
- `torch.ops.mslk.f8f8f16_rowwise` → FlyDSL (fp16 output)
- `torch.ops.mslk.f8f8bf16_rowwise_out` → FlyDSL (out-tensor variant)

**Benchmark: `FP8RowwisePreshuffleFlyDSL`** — new benchmark class in `bench/gemm/gemm_ops.py` targeting `AMD_GFX950`, gated on `is_flydsl_available()`.

## Test Plan

```bash
# Correctness test
pytest test/gemm/gemm_test.py -k FlyDSLPreshuffleGemmTest

# Benchmark
python bench/gemm/gemm_bench.py --op FP8RowwisePreshuffleFlyDSL --M 4096 --N 7424 --K 8192
```

### FlyDSL vs CK bpreshuffle benchmark (gfx950 MI350)

Both backends use the same bpreshuffle weight layout. CK forced to CK-only kernel path. All numbers are mean ± stddev over 3 independent runs.

#### GPU kernel time (rocprof --stats, 10 iters per run)

[...]

PR #456, opened 2026-07-27
Add FlyDSL grouped groupwise FP8 GEMM for ROCm

aryaman-gupta:
Adds a FlyDSL implementation of the grouped groupwise FP8 GEMM for ROCm and makes it the ROCm backend for `mslk::f8f8bf16_groupwise_grouped`.

Two ops share one kernel, selected by a compile-time `b_preshuffled` flag:

- **`mslk::f8f8bf16_groupwise_grouped`** — plain row-major `[G, N, K]` weights.
- **`mslk::f8f8bf16_groupwise_grouped_preshuffle`** — new ROCm-only sibling taking weights already swizzled into the MFMA B layout by `mslk.quantize.shuffle.preshuffle_b_mfma`. Callers shuffle once at load time and the op does no shuffling.

Both require `N` and `K` to be multiples of 128, matching the weight scale-block granularity. Activation scales are consumed in the per-group block layout produced by `quantize_fp8_group(m_sizes=...)`.

## Performance

CUDA-graph replay latency in µs on gfx950 (MI350X), G=8. **Both implementations are tuned**: Triton through its own `@triton.autotune`, FlyDSL through MSLK autotune over its full candidate tile set.

The last two columns are the latency reduction relative to Triton, `(t_triton − t_flydsl) / t_triton × 100`, so **higher is better** and a negative value is a slowdown.

Representative rows (two per shape) from a 50-row sweep. Preshuffle is faster on 45 of 50 shapes, median **+28%**, range −25% to +66%.

| totM | N | K | triton µs | plain µs | preshuffle µs | plain | preshuffle |
|-----:|----:|-----:|----------:|---------:|--------------:|------:|-----------:|
| 32 | 1280 | 8192 | 42.9 | 65.2 | 53.5 | −52% | −25% |
| 128 | 1280 | 8192 | 82.0 | 66.0 | 54.5 | +20% | +34% |
| 1024 | 2048 | 7168 | 99.4 | 62.2 | 53.0 | +37% | +47% |
| 16384 | 2048 | 7168 | 699.5 | 470.9 | 405.4 | +33% | +42% |
| 1024 | 7168 | 2304 | 74.1 | 42.8 | 43.5 | +42% | +41% |
| 16384 | 7168 | 2304 | 623.6 | 547.0 | 482.6 | +12% | +23% |
| 32 | 7168 | 8192 | 111.8 | 119.3 | 97.0 | −7% | +13% |
| 128 | 7168 | 8192 | 156.8 | 116.2 | 102.0 | +26% | +35% |
| 1 | 8192 | 1024 | 20.7 | 13.4 | 12.0 | +35% | +42% |
| 128 | 8192 | 1024 | 29.6 | 19.8 | 19.9 | +33% | +33% |
| 32 | 8192 | 3584 | 46.2 | 41.9 | 39.9 | +9% | +14% |
| 128 | 8192 | 3584 | 62.8 | 43.4 | 41.0 | +31% | +35% |
| 32 | 13312 | 6656 | 158.8 | 154.0 | 146.1 | +3% | +8% |
| 128 | 13312 | 6656 | 231.8 | 157.2 | 148.6 | +32% | +36% |
| 32 | 13312 | 16384 | 373.2 | 453.5 | 367.2 | −22% | +2% |
| 128 | 13312 | 16384 | 549.0 | 463.6 | 380.7 | +16% | +31% |
| 32 | 16384 | 6656 | 266.7 | 167.1 | 162.6 | +37% | +39% |
| 128 | 16384 | 6656 | 507.5 | 186.2 | 174.2 | +63% | +66% |

FlyDSL is slower on `1280 × 8192` up to totM 96, where the output is narrow enough that it cannot produce as many tiles as Triton; it leads on every other shape measured.

Every shape is checked against a per-group bf16 reference before timing, and additionally against Triton: the two agree to within one bf16 ulp on every shape, and the residual against the reference is FP8 quantization error, identical for both (~3.5% of max|ref|).

## Tile selection

With `MSLK_AUTOTUNE_ENABLE` set, FlyDSL autotune benchmarks the candidate tiles on a cache miss and persists the winner to disk, keyed on `(nextPow2(total_M), N, K, b_preshuffled)`. This follows the CUTLASS precedent in `f8f8bf16_groupwise_grouped.cu`, including the env gate: unset — the default, and what CI uses — a fixed tile is used and nothing is benchmarked.

Benchmarking cannot run inside a graph capture, so a shape must be warmed before it is captured. Everything else about dispatch is capture-safe: the grid extent is derived from static shapes alone, nothing on the host reads tensor contents, and no helper kernels are launched, so eager and captured dispatch issue the same single kernel.

## Architecture support

gfx950 uses the wide `mfma_scale_f32_16x16x128_f8f6f4` with a neutral E8M0 scale and applies the FP32 scales in software. gfx942 lacks that instruction and takes a narrow `16x16x32` path. LDS budgets are checked per architecture (64 KiB on gfx942, 160 KiB on gfx950), so a tile that fits on MI350 is rejected at compile time on MI300 rather than failing in the backend.

Correctness has been verified on gfx950 (MI350) and gfx942 (MI300).

## Testing

`test/gemm/gemm_test.py::test_f8f8bf16_groupwise_grouped` covers both variants across four group splits, including uneven and non-tile-aligned ones, and requires FlyDSL on ROCm.

The tolerance there is tightened from `atol=8e-2` to `1e-2`. Outputs on those shapes are O(1e-2), so the previous bound sat several times above the signal and would pass regardless of which scales the kernel applied.

## Notes for reviewers

ROCm CI already triggers on this change; `mslk/gemm/flydsl/**` is added to its path filter so later changes confined to the op wrappers do not skip it.

The FlyDSL modules vendored here are also vendored, at the same path, by #434 and #444. If either lands first, this PR will be refactored to drop the repeated files.

`mslk/flydsl/kernels/{common,mma}/` and `gemm/fp8_gemm_utils.py` are vendored unmodified from FlyDSL and are best reviewed as an import. The kernel itself is `mslk/flydsl/kernels/gemm/grouped_gemm_blockscale_{contiguous,common}.py`, and the op wrapper is `mslk/gemm/flydsl/fp8_groupwise_grouped_gemm.py`.


q10:
@aryaman-gupta could you rebase this PR on latest main?

PR #471, opened 2026-08-06
Back the FP8 rowwise grouped GEMM ops with FlyDSL on ROCm

aryaman-gupta:
## Summary

Implements the FP8 **rowwise**-scaled grouped GEMM entry points with FlyDSL and binds them on
ROCm, extending the grouped kernel added for the groupwise op in #470. On ROCm these ops are now
served by FlyDSL rather than CK; on CUDA nothing changes.

| Op | Group layout | Status |
|---|---|---|
| `f8f8bf16_rowwise_grouped_stacked` | groups packed along M, `[G]` int64 row counts | FlyDSL |
| `f8f8bf16_rowwise_grouped_dynamic` | fixed `[G, M, K]` slabs, runtime valid-row count | FlyDSL |
| `f8f8bf16_rowwise_grouped_mm` | 2D-3D (M), 3D-3D (batched), 3D-2D (N), 2D-2D (K) | FlyDSL |
| `..._stacked_preshuffle`, `..._dynamic_preshuffle`, `..._mm_preshuffle` | as above, MFMA-preshuffled B | FlyDSL (new schemas) |

`f8f8bf16_rowwise_grouped` and `_cat` are left on CK: both are marked
`UNSUPPORTED AND DEPRECATED` in CK's own source, have no in-repo callers and no tests.

## Design notes

**One kernel, not six.** The group geometry is a compile-time `layout` enum — `sizes`, `offsets`,
`padded`, `batched`, `n_offsets`, `k_offsets`. Only the resolution step differs between them; the
loaders, K loop and epilogue are shared. `n_offsets` (groups divide N) and `k_offsets` (groups
divide K) resolve the group from the N-block index and from the grid z axis respectively, rather
than from a row-count scan.

**Scaling scheme is a compile-time flag.** Rowwise carries one scale per row of A and per column
of B, both constant along K, so they factor out of the reduction and apply once in the epilogue
instead of per scale block inside the K loop. This also frees the tile from scale-block alignment,
making `tile_n = 64` available, which block scaling cannot express.

**The K loop is rolled by default.** Unrolling folds the tile index into every address, but traced
IR, compile time and code size then grow linearly with K: at K=8192 the plain loop is 5271 lines of
ISA and 12.9 s to compile unrolled, against 990 lines and 1.5 s rolled, with bitwise-identical
output. Both loop forms share one body. `k_offsets` is always rolled, since a group's K length is
a runtime value.

**Preshuffled B is offered only where a group owns a whole `[N, K]`.** The MFMA swizzle interleaves
N and K across the whole matrix, so a group boundary inside either axis falls inside the swizzle.
`_mm_preshuffle` therefore serves the 2D-3D and 3D-3D ranks and rejects 3D-2D and 2D-2D rather than
silently ignoring weights the caller has already shuffled.

**Tile selection is opt-in.** Tuning runs only under `MSLK_AUTOTUNE_ENABLE`, matching the CK and
CUTLASS paths; otherwise a fixed default tile is used with no benchmarking. `mslk/flydsl/autotune.py`
is generic and carries nothing specific to this kernel.

## Performance

FlyDSL vs CK, CUDA-graph replay, G=8, gfx950, idle GPU. Figures are the reduction in kernel time
against CK,

```
(ck_time - fly_time) / ck_time * 100
```

so a positive number is the share of CK's time saved and a negative one is the share added.
**59 of 72 shapes are faster.**

| Variant | Shapes faster | Median | Worst | Best |
|---|---|---|---|---|
| `_mm` 2D-3D | 18/24 | +10.3% | -16.7% | +39.8% |
| `_mm` 3D-3D | 11/12 | +11.7% | -12.9% | +20.4% |
| `_mm` 3D-2D | 8/12 | +12.0% | -12.2% | +35.9% |
| `_mm` 2D-2D | 12/12 | +22.4% | +0.9% | +41.3% |
| `_dynamic` | 10/12 | +30.5% | -38.5% | +60.2% |

Two caveats belong with these numbers:

- **CK runs on its heuristic; FlyDSL is autotuned.** `MSLK_AUTOTUNE_ENABLE` is read by both, and
  CK's tuning path raises `Kernels to tune over is empty` for these entry points, so the benchmark
  toggles it per side. Each is at its best *available* setting, not the same setting.
- **The largest cluster of slower shapes is narrow N.** Six of the thirteen are `1280x8192`, where
  there are too few N-blocks to fill the machine; the same weakness is reported in #470 against
  Triton, so it reproduces across two kernels and two baselines. Most of the remainder are
  total_M=1 at large N, where launch overhead dominates.

Both sides are timed after a 0.5 s clock ramp. Without it, a short window on an idle GPU measures
the clock ramp as much as the kernel — about 12% on these shapes — and whichever side is timed
second inherits the other's warm-up.

<details>
<summary>Per-shape detail</summary>

`_mm` 2D-3D — M is total rows across all groups

| Family | M | N×K | CK µs | FlyDSL µs | Time saved |
|---|---|---|---|---|---|
| ds_v3 | 1 | 2048×7168 | 30.1 | 25.5 | +15.3% |
| ds_v3 | 128 | 2048×7168 | 31.8 | 33.7 | -6.0% |
| ds_v3 | 1024 | 2048×7168 | 47.6 | 39.8 | +16.4% |
| ds_v3 | 4096 | 2048×7168 | 99.5 | 89.8 | +9.7% |
| ds_v3 | 1 | 7168×2304 | 19.7 | 13.1 | +33.5% |
| ds_v3 | 128 | 7168×2304 | 44.0 | 26.5 | +39.8% |
| ds_v3 | 1024 | 7168×2304 | 39.1 | 36.3 | +7.2% |
| ds_v3 | 4096 | 7168×2304 | 105.9 | 103.0 | +2.7% |
| llama3_70b | 1 | 1280×8192 | 32.5 | 28.0 | +13.8% |
| llama3_70b | 128 | 1280×8192 | 33.3 | 35.9 | -7.8% |
| llama3_70b | 1024 | 1280×8192 | 50.8 | 44.4 | +12.6% |
| llama3_70b | 4096 | 1280×8192 | 81.9 | 95.6 | -16.7% |
| llama3_70b | 1 | 8192×1024 | 16.0 | 9.9 | +38.1% |
| llama3_70b | 128 | 8192×1024 | 22.1 | 18.1 | +18.1% |
| llama3_70b | 1024 | 8192×1024 | 27.9 | 23.9 | +14.3% |
| llama3_70b | 4096 | 8192×1024 | 69.2 | 55.7 | +19.5% |
| llama3_405b | 1 | 13312×16384 | 56.6 | 58.9 | -4.1% |
| llama3_405b | 128 | 13312×16384 | 352.9 | 362.2 | -2.6% |
| llama3_405b | 1024 | 13312×16384 | 479.5 | 430.0 | +10.3% |
| llama3_405b | 4096 | 13312×16384 | 1295.6 | 1271.5 | +1.9% |
| llama3_405b | 1 | 16384×6656 | 28.3 | 31.4 | -11.0% |
| llama3_405b | 128 | 16384×6656 | 173.1 | 162.7 | +6.0% |
| llama3_405b | 1024 | 16384×6656 | 229.4 | 211.1 | +8.0% |
| llama3_405b | 4096 | 16384×6656 | 664.0 | 589.2 | +11.3% |

`_dynamic` — M is the per-group slab height, 75% of rows valid

| Family | M | N×K | CK µs | FlyDSL µs | Time saved |
|---|---|---|---|---|---|
| ds_v3 | 128 | 2048×7168 | 74.7 | 44.1 | +41.0% |
| ds_v3 | 512 | 2048×7168 | 96.1 | 86.6 | +9.9% |
| ds_v3 | 128 | 7168×2304 | 104.6 | 41.6 | +60.2% |
| ds_v3 | 512 | 7168×2304 | 123.1 | 99.1 | +19.5% |
| llama3_70b | 128 | 1280×8192 | 48.3 | 48.6 | -0.6% |
| llama3_70b | 512 | 1280×8192 | 69.6 | 96.4 | -38.5% |
| llama3_70b | 128 | 8192×1024 | 47.6 | 29.2 | +38.7% |
| llama3_70b | 512 | 8192×1024 | 76.7 | 61.7 | +19.6% |
| llama3_405b | 128 | 13312×16384 | 630.0 | 443.6 | +29.6% |
| llama3_405b | 512 | 13312×16384 | 2072.9 | 1097.2 | +47.1% |
| llama3_405b | 128 | 16384×6656 | 313.0 | 217.4 | +30.5% |
| llama3_405b | 512 | 16384×6656 | 890.3 | 551.2 | +38.1% |

`_mm` 3D-3D, 3D-2D and 2D-2D follow the same pattern; medians are in the summary table.

</details>

## Correctness

`test/gemm/gemm_test.py -k "grouped or batched_gemm"`: **92 passed, 67 skipped, 0 failed** on
gfx950. The existing upstream contract tests for `_stacked`, `_dynamic` and all four `_mm` rank
combinations now exercise FlyDSL rather than CK, so they are the primary coverage.

Added tests cover the boundaries the existing cases did not reach:

- group extents that do not land on a tile boundary, along K (2D-2D), N (3D-2D) and the slab height
  (3D-3D)
- the preshuffled grouped `mm`, and rejection of the two ranks that cannot take preshuffled weights
- block scaling refusing a K or N that covers only part of a scale block

The group-K boundary test gives alternate groups constant, single-signed data. A read that runs
past a group's K end then accumulates coherently rather than cancelling; with random data on both
sides the resulting error is the same order as the FP8 quantisation noise and sits inside this
file's comparison tolerance, so the test would pass whether or not the boundary is respected.

Across the benchmark, `max|fly − ck|` is at most 0.0039 on every row.

Unit tests pass on MI350 (gfx950) and MI300 (gfx942).

## Notes for reviewers

- **`gemm_ops.cpp` frees the CK slots on ROCm** for `_stacked`, `_dynamic` and `_mm`, so the Python
  binding is the only implementation there rather than an override of one. A consequence is that a
  ROCm build without FlyDSL has no implementation for these ops; this already held for `_stacked`.
  CK continues to serve every one of them on CUDA.
- **This requires a rebuild.** Three schemas are new, and the registration guards only take effect
  once `gemm_ops.cpp` is recompiled.
- **Two kernel modules are renamed.** `grouped_gemm_blockscale_{common,contiguous}.py` become
  `fp8_grouped_gemm{_common,}.py`: the kernel serves both scaling schemes and six layouts, so
  neither "blockscale" nor "contiguous" described it. No behaviour change — the compile-cache key
  is built from its own strings and is unaffected.
- **Per-group contracts that cannot be checked on the host.** The offsets live on the device, so
  like CK these are documented rather than asserted: group K a multiple of 16 (2D-2D) and group N a
  multiple of 8 (3D-2D). Measured against a dequantised reference, CK and FlyDSL are equally
  inaccurate when group K is not a multiple of 16, so this is not a change in what works. For
  ragged N, the epilogue drops the store that would cross a group boundary, leaving those columns
  unwritten, rather than overrunning into the next group.
- **ROCm CI now triggers on `mslk/gemm/flydsl/**`**, which the path filter did not previously cover.

### DIFF
--- a/.github/workflows/mslk_ci_rocm.yml
+++ b/.github/workflows/mslk_ci_rocm.yml
@@ -41,6 +41,8 @@ on:
       - 'mslk/attention/fmha/flydsl_decoder.py'
       - 'mslk/attention/fmha/flydsl_splitk.py'
       - 'test/attention/fmha/**'
+      # FlyDSL GEMM ops
+      - 'mslk/gemm/flydsl/**'
       # GEMM tests
       - 'test/gemm/gemm_test.py'
       # AMD/ROCm Triton GEMM kernels
--- a/bench/gemm/gemm_ops.py
+++ b/bench/gemm/gemm_ops.py
@@ -24,6 +24,7 @@ from mslk.quantize.shuffle import (
     ck_preshuffle,
     int4_row_quantize_zp,
     pack_int4,
+    preshuffle_b_mfma,
     quantize_int4_preshuffle,
 )
 from mslk.quantize.triton.fp4_quantize import (
@@ -1100,6 +1101,7 @@ class FP8RowwiseGrouped(GemmOpBase):
             Accelerator.NVIDIA_SM100,
             Accelerator.NVIDIA_SM103,
             Accelerator.AMD_GFX942,
+            Accelerator.AMD_GFX950,
         }
 
     @property
@@ -1129,7 +1131,7 @@ class FP8RowwiseGrouped2D3D(FP8RowwiseGrouped):
 
     @property
     def supported_accelerators(self) -> set[Accelerator]:
-        return {Accelerator.AMD_GFX942}
+        return {Accelerator.AMD_GFX942, Accelerator.AMD_GFX950}
 
     @property
     def supported_gemm_types(self) -> set[GemmType]:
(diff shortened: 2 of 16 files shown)


jwfromm:
Impressive speedups! Ill take a look at this today so we can get it merged :)

jwfromm:
It would be great if you could also add wrappers for these new variants to `https://github.com/meta-pytorch/MSLK/blob/main/bench/gemm/gemm_ops.py` as it will make is much easier for folks to reproduce the numerics and perf of the ops.

aryaman-gupta:
> It would be great if you could also add wrappers for these new variants to `https://github.com/meta-pytorch/MSLK/blob/main/bench/gemm/gemm_ops.py` as it will make is much easier for folks to reproduce the numerics and perf of the ops.

Sure @jwfromm. Please give me until tomorrow, I will do this

aryaman-gupta:
> It would be great if you could also add wrappers for these new variants to `https://github.com/meta-pytorch/MSLK/blob/main/bench/gemm/gemm_ops.py` as it will make is much easier for folks to reproduce the numerics and perf of the ops.

Hi @jwfromm, done in https://github.com/meta-pytorch/MSLK/pull/471/commits/9534de0a768ad7a0e16ab4559a6c2be46bad1330. Could you please re-import?

cthi:
@aryaman-gupta  can you rebase this, we had some CI breakage, thanks.

aryaman-gupta:
> @aryaman-gupta can you rebase this, we had some CI breakage, thanks.

Sure @cthi, done

jwfromm:
Thanks for the updates @aryaman-gupta, ill reimport this and try to get it landed asap.

jwfromm:
@aryaman-gupta can you check the errors? Everything else is passing internally so we'd just want to make sure the `E   ModuleNotFoundError: No module named 'mslk.flydsl.kernels.gemm.grouped_gemm_blockscale_common'` error is expected.

aryaman-gupta:
@jwfromm Not expected — thanks for catching it.

This branch renames `grouped_gemm_blockscale_common.py` to `fp8_grouped_gemm_common.py` (the module holds a fair amount of non-blockscale code). `gemm_blockscale.py` landed after the branch point, so we didn't see that it imports 21 helpers from that module — your internal main already had the blockwise kernel, which is why the import surfaced it before our tree did.

Fixed in 51add66, and the branch is rebased onto latest main.

jwfromm:
Great, i think we're good to merge. I'll click the button as soon as internal CI finishes. Thank you @aryaman-gupta!

PR #463, opened 2026-07-29
[ROCm] Add FlyDSL paged-attention decode backend (dense + fp8) with benchmark

avbokovoy:
UPD (08.13): As per request, splitting this PR into 2:
https://github.com/meta-pytorch/MSLK/pull/483
https://github.com/meta-pytorch/MSLK/pull/484

## Summary

Replaces the CK paged-attention decode operators with a FlyDSL backend and adds a native-fp8 decode path, wired into the existing fmha op registry (flydsl_decoder / flydsl_splitk) with backward-compatible op aliases. FlyDSL becomes the sole decode backend across gfx942 and gfx950; the CK operator path is removed.

## Kernels

Dense f16/bf16 decode uses three kernels selected by shape: pa_decode_gfx950 (head-packed MFMA + double-buffered wide V load, ds_read_tr16_b64; gfx950, GQA ratio 1..16), pa_decode_gfx950_coop (gfx950 cooperative-DMA for ratios that cannot head-pack), and pa_decode_generic (arch-generic per-warp fallback for gfx942 and off-gfx950). pa_decode_fp8 adds native fp8 (e4m3fn + symmetric per-token scale) paged decode covering MQA/GQA, D=128/256, and any context length; it is opt-in per call via Inputs.quantize_kv_to_fp8 and is CUDA-graph capturable. All kernels (dense, split-K reduce, and fp8) are AOT-registered and precompiled into the bundled cache.

## Performance

Measured on gfx950 (MI350X). Dense f16/bf16: FlyDSL is 1.6-10x faster than the old CK decoder (CK uses no matrix cores) and within 1.05-1.37x of Triton (near parity; Triton is marginally faster on raw kernel time). Native fp8: FlyDSL is 1.8-2.5x faster than Triton fp8 and also beats Triton dense on most shapes. Correctness is verified against a torch reference for all backends.

## Other changes

Fixes the Triton fp8 decode to use OCP e4m3fn on gfx950 (was hardcoded fnuz, which mis-decoded gfx950 caches). Aligns FlyDSL helper imports with the mslk.flydsl package layout. Adds bench/attn/decoder_bench.py (FlyDSL vs Triton, dense and fp8) with eager timing via the shared do_bench, real CUDA-graph timing with empty-graph detection, and per-shape subprocess isolation. Adds ROCm CI path triggers for the decode backend and its tests.

## Test plan

- [x] test_flydsl_fp8_decoder (108 passed on gfx950)
[...]

PR #467, opened 2026-08-04
[feat] add flyDSL bwd FMHA kernel 

amd-weisun:
# Add FlyDSL FMHA backward kernel (gfx950 + gfx942 fallback)

## Summary

Adds a FlyDSL FMHA backward implementation, registered as `flydsl.BwOp` (an
opt-in `AttentionBwOpBase`, following the same pattern as `flash.BwOp` /
`flash3.BwOp`). It is **not** wired into `dispatch.py`'s live `_dispatch_bw()`
priority list (which still uses `ck.BwOp` on ROCm) — this PR adds the op and
its test coverage, not a change to production routing.

Two kernels ship together:
- **gfx950 (CDNA4/MI350X):** fused dQ+dV+dK kernel
  optimized for this architecture (hardware LDS transpose, XOR-swizzled
  layouts, register-resident K/V/KT, software-pipelined prefetch — see below).
- **gfx942 (CDNA3) fallback:** a 32x32 MFMA kernel (`fmha_bwd_mfma.py`) used
  when gfx950 isn't available.

### Files changed

| File | Description |
|------|-------------|
| `mslk/attention/flydsl/fmha_bwd_mfma_gfx950.py` | New gfx950 kernel (1710 lines) |
| `mslk/attention/flydsl/fmha_bwd_mfma.py` | New gfx942 fallback kernel (32x32 MFMA) |
| `mslk/attention/flydsl/fmha_bwd_preprocess.py` | New D_vec preprocess kernel |
| `mslk/attention/flydsl/fmha_bwd_convert_dq.py` | New dQ f32→output-dtype convert kernel |
| `mslk/attention/fmha/flydsl.py` | New `flydsl.BwOp`: gfx950 routes to the new kernel with `ck_scope_dvdk=True`; gfx942 falls back to `fmha_bwd_mfma.py` |
| `mslk/attention/fmha/__init__.py` | Register `flydsl.BwOp` in `ALL_BW_OPS` on ROCm (test enumeration only) |
| `test/attention/fmha/test_backward.py` | Add `flydsl.BwOp` to test_backward + test_backward_gqa; 3 new negative-path tests |

### Test plan

- [x] `test_backward.py -k flydsl`: 316 passed, 0 failed
- [x] `test_backward_gqa` (flydsl.BwOp): 2 passed, 2 skipped (bf16 precision skip, matching CK's own)
- [x] Kernel-level correctness suite (direct calls to `compile_fmha_bwd_dqdkdv_mfma_gfx950`,
[...]

PR #483, opened 2026-08-13
[ROCm] Flydsl decode backend

avbokovoy:
As per request from https://github.com/meta-pytorch/MSLK/pull/463#pullrequestreview-4906645518


bottler:
Seems odd that this code is in mslk/attention/fmha/flydsl but the recent backward flydsl code #467 is in mslk/attention/flydsl.

cthi:
Yeah I agree with @bottler , can AMD align on a common directory structure for attn? e.g. where kernels should go, regardless of how we integrate it into fmha. cc @liligwu

avbokovoy:
@bottler @cthi Rebased and restructured the branch to align with current Flash/CUTLASS backends (Triton uses different file structure). Could you verify if it complies with repo requirements?

bottler:
This makes mslk/attention/flydsl contain two independent groups of files implementing different backends, should they be split into subdirectories? I think there was a tiny bit of shared code in the previous version of this PR, but there's none now.

avbokovoy:
> This makes mslk/attention/flydsl contain two independent groups of files implementing different backends, should they be split into subdirectories? I think there was a tiny bit of shared code in the previous version of this PR, but there's none now.

@bottler I don't think that we should take into consideration shared code while making the decision on subdirectories split. If you would ask me, I lean towards more radical repo-wide restructuring, however I just want to stick with the solution that will be clear and reasonable for this repo and doesn't break anything. So, we definitely can split the algorithms into subdirectories and I think it's reasonable, however I believe it should be a follow-up PR after this one is merged. Wdyt?

bottler:
Let's do _something_ about folder structure in this PR. It just seems unfriendly to have two unrelated groups of files in the same directory (unless their names were grouped I guess). Easiest change would be to make this PR put its file in a subdir called e.g. `decode`. Then there's no clash with #485.

avbokovoy:
> Let's do _something_ about folder structure in this PR. It just seems unfriendly to have two unrelated groups of files in the same directory (unless their names were grouped I guess). Easiest change would be to make this PR put its file in a subdir called e.g. `decode`. Then there's no clash with #485.

Done restructuring in https://github.com/meta-pytorch/MSLK/pull/483/commits/ae94844b82ff5cff403fa10355af06d4f6f4d232

PR #485, opened 2026-08-13
[ROCm] FlyDSL forward attention op

avbokovoy:
Adds `flydsl_forward.FwOp`, a FlyDSL flash-attention forward operator that wraps `mslk.attention.flydsl.flydsl_flash_attn_func` behind the `AttentionFwOpBase` interface as a drop-in for `ck.FwOp`, together with the kernel-side features it needs (split-K, sliding-window tile skipping, in-kernel gather for gappy/padded/paged KV, dropout, and LSE on the generic path).

## Summary

- **dtypes / head dim**: bf16 and f16; any `head_dim` in `[1, 256]` (dims the kernel does not build natively are zero-padded to the next of `{64, 128, 256}`, with the native softmax scale pinned so padding does not change the result).
- **formats**: BMHK and BMGHK (GQA); the G axis is folded into H the same way `ck.FwOp` does.
- **attn_bias**: full parity with CK, all 18 xformers types -- causal / non-causal self-attention, additive tensor bias, sliding-window (local) masks, varlen (`BlockDiagonal*`), padded/gappy keys, and paged KV. Paged KV accepts any page_size that is a multiple of 64 (remapped to 64-row sub-pages); paged-gappy uses an in-kernel per-row page lookup.
- **LSE** output supported (including the generic paged path).
- **dropout** matches CK's philox exactly (via `torch.ops.xformers._ck_rand_uniform`) and records `[seed, offset]` in `ctx.rng_state` so `ck.BwOp` reproduces the mask.
- **split-K** on the generic kernel for short-Q dense/causal shapes, gated by a Python port of CK's occupancy heuristic; a combine kernel reduces the per-split partials into the final O and LSE.
- **CUDA-graph**: every supported bias path is graph-capturable (no device-to-host syncs on the capture path).
- Registered in `ALL_FW_OPS` and the ROCm dispatch priority list after `ck.FwOp`: it does not displace the default (auto-dispatch still selects `ckF`), but is available when CK declines or when the op is selected explicitly.
- Backward falls back to `ck.BwOp` (no FlyDSL backward).

## Behavior change / scope (not limited to the new op)

  - N128 kernel selection now applies to non-causal D=128 (f16/bf16), not just causal. Previously the N128 generic path was gated on causal; that gate was removed, so existing non-causal D=128 callers of the
  generic kernel now select N128 instead of N32. N128 is faster for D=128, but this changes the kernel used for callers unrelated to flydslF. (See flash_attn_utils.py:_make_flash_attn_generic_traits.)

## Correctness

A full sweep runs `flydsl_forward.FwOp` and `ck.FwOp` on identical inputs across the whole CK-supported matrix (all 18 attn_bias types, bf16 and f16, BMHK and BMGHK/GQA, head dims 64/96/128/256, short and long q, varied batch), comparing wherever each op accepts the case. Result: 222 cases, 0 failures, 0 skips. Every case matches CK within the reduced-precision floor (worst bf16 relerr 0.019, f16 far lower), with no NaN. Zero skips means the op accepts every case CK accepts.

## Performance

CUDA-graph replay timing on gfx950 (MI355X), FlyDSL vs `ck.FwOp` (the op this replaces). `fly/ck < 1` means FlyDSL is faster. Inputs use the test suite's scale-3 tensors, so error is reported relative to CK output magnitude (bf16/f16 floor is about 0.004 to 0.02). Full per-shape tables in `BENCHMARKS.md`.

Representative bf16 results:

| bias type        | shape (B,q,kv,D)   | FlyDSL us | CK us  | fly/ck |
|------------------|--------------------|-----------|--------|--------|
| none             | 1, 4096, 4096, 128 | 126.0     | 234.0  | 0.54   |
| causal           | 1, 4096, 4096, 128 | 111.0     | 216.9  | 0.51   |
| window           | 1, 2048, 2048, 128 | 13.4      | 22.7   | 0.59   |
| varlen_causal    | 8, 1024, 1024, 128 | 113.7     | 227.1  | 0.50   |
| varlen_causal_br | 4, 512, 512, 128   | 36.6      | 84.6   | 0.43   |
| gappy_causal     | 4, 256, 512, 128   | 43.9      | 83.8   | 0.52   |
| paged            | 4, 256, 512, 128   | 37.5      | 47.2   | 0.79   |
| paged_gappy      | 4, 256, 512, 128   | 42.7      | 51.0   | 0.84   |
| tensorbias       | 1, 2048, 2048, 128 | 114.1     | 106.1  | 1.08   |

Representative f16 results (same shapes):

| bias type        | shape (B,q,kv,D)   | FlyDSL us | CK us  | fly/ck |
|------------------|--------------------|-----------|--------|--------|
| none             | 1, 4096, 4096, 128 | 127.0     | 260.2  | 0.49   |
| causal           | 1, 4096, 4096, 128 | 105.3     | 209.2  | 0.50   |
| window           | 1, 2048, 2048, 128 | 15.2      | 25.0   | 0.61   |
| varlen_causal    | 8, 1024, 1024, 128 | 158.3     | 211.2  | 0.75   |
| varlen_causal_br | 4, 512, 512, 128   | 32.0      | 70.8   | 0.45   |
| gappy_causal     | 4, 256, 512, 128   | 35.2      | 62.7   | 0.56   |
| paged            | 4, 256, 512, 128   | 35.9      | 40.3   | 0.89   |
| paged_gappy      | 4, 256, 512, 128   | 42.6      | 49.0   | 0.87   |
| tensorbias       | 1, 2048, 2048, 128 | 121.7     | 110.7  | 1.10   |

FlyDSL is faster on the large majority of type/shape/format combinations in both dtypes, with the biggest wins on long-Q prefill (up to ~2x at S=4096), sliding-window at large S (~1.6x), and cross-length varlen causal (up to 2x). f16 tracks bf16 closely and is numerically tighter (relerr ~0.0005 to 0.0006 vs CK, since it skips the bf16 mantissa floor). GQA tracks BMHK. The few near-parity cases are consistent across both dtypes: additive tensor bias at S=2048 (intrinsic full-plane bias streaming, ~1.05 to 1.10x), small-batch varlen_causal (launch-bound, ~1.10x), and paged GQA (~1.07 to 1.09x).

## Testing

- `test/attention/fmha/test_forward_flydsl.py` -- stock forward suite scoped to the op (skips declined cases; xfails the documented scale-3 multi-tile precision floor).
- `test/attention/fmha/test_dropout_flydsl.py` -- philox parity, reproducibility, and forward+backward vs CK.
- `test/attention/fmha/test_mem_eff_attention.py` -- LSE and forward coverage for the op (exercised via `ALL_FW_OPS`, including GQA and all bias types).

## Known limitations

- No FlyDSL backward; backward uses `ck.BwOp`.


mslk/attention/fmha/flydsl_forward.py
```diff
@@ -0,0 +1,543 @@
...
+
+    # Dropout (dense): reuses CK's philox mask and records [seed, offset] in
+    # ctx.rng_state so ck.BwOp reproduces it. See _apply_bmhk.
+    SUPPORTS_DROPOUT = True
+    SUPPORTS_CUSTOM_SCALE = True
+    SUPPORTS_DIFFERENT_VALUE_EMBED = False
+    SUPPORTS_BMGHK = True
+    SUPPORTS_PARTIAL = False
+    # Varlen LSE is returned padded per-batch as [B, H, max_seqlen_q], not packed.
+    VARLEN_LSE_PACKED = False
```
q10: Set `VARLEN_LSE_PACKED` to match the backward operator.

`FwOp` sets it to `False`, but `ck.BwOp` uses `True`. The exported operator rejects multi-sequence `BlockDiagonal*` inputs. Automatic dispatch also cannot select a backward operator for variable-length inputs that require gradients.

Emit packed LSE and set this value to `True`. Alternatively, reject variable-length inputs that require gradients in `not_supported_reasons`. Add an end-to-end test for the exported forward/backward pair.

mslk/attention/fmha/flydsl_forward.py
```diff
@@ -0,0 +1,543 @@
...
+            value=value,
+            attn_bias=bias_flat,
+            p=inp.p,
+            scale=inp.scale,
+            output_dtype=inp.output_dtype,
+        )
+        out, ctx = cls._apply_bmhk(flat, needs_gradient)
+        out = out.unflatten(2, (G, Hq))
+        if ctx is not None:
+            ctx = Context(lse=ctx.lse.unflatten(1, (G, Hq)), out=out, op_bw=ctx.op_bw)
```
q10: Preserve `rng_state` on the BMGHK path.

This code creates a new `Context` and discards the dropout state from `_apply_bmhk`. `ck.BwOp` then receives `rng_state=None` for 5-D dropout inputs and raises `NotImplementedError`.

Update `ctx.lse` and `ctx.out` on the existing context. Add a BMGHK dropout test with gradients.

mslk/attention/fmha/flydsl_forward.py
```diff
@@ -0,0 +1,543 @@
...
+            Sq = inp.query.shape[1]
+            H = inp.query.shape[2]
+            Skv = inp.key.shape[1]
+            gen = torch.cuda.default_generators[inp.query.device.index]
+            seed = int(gen.initial_seed())
+            offset = int(gen.get_offset())
+            # _ck_rand_uniform advances the generator like CK's fused forward and
+            # returns a [B, H, Sq, Skv] uint8 randval; keep <=> rand <= (1-p)*255.
+            pattern = torch.empty((B, H, Sq, Skv), device=inp.query.device)
+            rand_uniform = torch.ops.xformers._ck_rand_uniform(inp.p, pattern)
```
q10: Replace `_ck_rand_uniform` in this production dropout path.

- The kernel uses the device-0 default generator. On another device, backward regenerates a different mask.
- The operator synchronizes the HIP stream. The host generator reads also prevent graph capture.
- `pattern` defaults to fp32 and materializes `[B, H, Sq, Skv]`. It uses about 1.9 GB for `B=1, H=16, S=4096`.

Use a device-correct, asynchronous RNG. Pass dimensions instead of a full tensor. At minimum, allocate `pattern` as `uint8`.

mslk/attention/fmha/flydsl_forward.py
```diff
@@ -0,0 +1,543 @@
...
+        torch.float: 2e-5,
+        torch.half: 3e-3,
+        torch.bfloat16: 2e-2,
+    }
+
+    # 96 is non-native, to exercise head-dim padding.
+    _TEST_K: List[int] = [64, 96, 128, 256]
+
+    @classmethod
+    def not_supported_reasons(cls, d: Inputs) -> List[str]:
```
q10: Make `not_supported_reasons` match the interface restrictions. Otherwise, dispatch selects this operator and `apply()` fails at run time.

Reject these inputs before dispatch:

- native paged attention with `return_lse` on an unsupported route;
- non-gappy paged attention with a custom scale;
- Q, K, or V without the required last-dimension stride and alignment;
- tensor bias whose dtype differs from the query dtype.

The last case currently reinterprets fp32 bias data as bf16 or f16.

mslk/attention/flydsl/flash_attn_utils.py
```diff
@@ -2261,6 +2369,38 @@ def init_descriptors(self, Q, O):  # noqa: E741
             base_byte_offset=q_batch_byte_off,
         )
 
+    def init_bias(self, Bias, bias_stride_b, bias_stride_h, bias_stride_q):
```
q10, line 2403: Bound the bias and dropout loads.

These paths use raw pointers, so hardware does not clamp out-of-range indexes. Tail tiles can use `q_row >= Sq` or `kv_col >= Skv`. This causes an out-of-bounds read when a sequence length is not tile-aligned.

Use a bounded buffer resource, or clamp both indexes before the GEP. Also use Int64 for the plane offset if the ABI supports it.

q10 (commented):
This review contains the OSS-applicable findings from Meta review D115895891.

Please also address these test and scope items:

- Use more than one head in dropout tests. Test f16 and bf16.
- Add the 222-case parity sweep and benchmark source, or link reproducible results.
- Link a task for the dualwave f16 and cross-attention NaN defects. The new exclusions move existing users to the generic path.

The window mask, gappy page lookup, split-K layout, BMGHK folding, and cache key are correct. Meta-only BUCK and internal CI findings are intentionally omitted.

mslk/attention/flydsl/flash_attn_generic.py
```diff
@@ -742,17 +1000,35 @@ def _compile(
...
+                drop_t,
+                0,
+                0,
+                0,
                 fx.Stream(stream),
             )
 
     _launch.compile = _compile
 
     def _wrap_pad_mask_dispatch(_inner):
```
q10, line 1025: Pass `num_kv_splits` to both builders in `_wrap_pad_mask_dispatch`.

For non-causal dense attention, these builders receive the default value. The kernel runs without split-K, although the interface allocates a split-K workspace. The output is correct, but split-K is a silent no-op and the workspace is unused.

Apply the same fix to the m128/m256 branch.

mslk/attention/flydsl/flash_attn_interface.py
```diff
@@ -86,16 +86,84 @@ def _dense_generic_tile(
...
+    blocks = num_batches * num_heads * ceildiv(seqlen_q, mtile)
+    if blocks >= 0.9 * num_cu:
+        return 1
+
+    max_splits = min(max_splits, num_cu)
+    max_check = 1
+    while _generic_splitk_list(max_check) <= max_splits:
+        max_check += 1
+
+    num_splits = 2
```
q10: Nit: Initialize `num_splits` to `1`.

The reported return of `2` is unreachable because the earlier occupancy guard returns `1` when `num_cu == 1`. This small change matches the no-split default and prevents the warning from recurring.

mslk/attention/fmha/flydsl_forward.py
```diff
@@ -0,0 +1,543 @@
...
+
+        def _set_varlen_kw(cu_q, cu_kv, max_q, max_kv, cross):
+            kw["cu_seqlens_q"] = cu_q
+            kw["cu_seqlens_kv"] = cu_kv
+            kw["max_seqlen_q"] = max_q
+            kw["max_seqlen_kv"] = max_kv
+            kw["cross_seqlen"] = cross
+
+        # Contiguous cu_seqlens (prefix-0 + cumsum) from per-seq lengths.
+        def _cu_seqlens(seqlens, device):
```
q10: Nit: Remove `_cu_seqlens` and the other unused definitions: `_PAGED_KERNEL_PAGE_SIZE`, `_VARLEN_WINDOW_TYPES`, and `_PADDED_GAPPY_CAUSAL_TYPES`.

Also move the local `math` import to module scope. `_cu_seqlens` performs a host-to-device copy and is not graph-safe.

mslk/attention/flydsl/flash_attn_utils.py
```diff
@@ -2285,6 +2430,48 @@ def init_kv_bounds(self):
...
+                traits.NUM_KV_SPLITS
+            )
+            t0 = self.split_idx * chunk
+            t1 = t0 + chunk
+            lo = base_lower + t0 * step
+            hi = base_lower + t1 * step
+            self.kv_lower = (lo < self.kv_upper).select(lo, self.kv_upper)
+            self.kv_upper = (hi < self.kv_upper).select(hi, self.kv_upper)
+            # This split is empty when its window has no KV positions.
+            self.split_empty = self.kv_lower >= self.kv_upper
```
q10, line 2508: Store a literal `-1e30` for an empty split, as the dualwave path does.

An empty split starts with `m_final = -inf`. Under `fm_fast`, the multiplication can become poison before `fmax` clamps it. A poisoned value can become `m_max` and reduce the final output to zero.

Alternatively, exclude entries with `l_i == 0` from `reduce_m_max`. Remove `self.split_empty` if the fix does not use it.

test/attention/fmha/test_forward_flydsl.py
```diff
@@ -0,0 +1,51 @@
...
+    eff_kv = max(Mq, Mkv)
+    try:
+        _stock_test_forward(opFW_device_dtype_biasT_B_Mq_Mkv_H_K_Kv_packed_fmt)
+    except ValueError as e:
+        # The op declines unsupported cases; the stock sweep force-feeds all of
+        # them, so treat a decline as a skip.
+        if "does not support inputs" in str(e) or "is not supported" in str(e):
+            pytest.skip("flydslF declined (not_supported_reasons)")
+        raise
+    except AssertionError:
```
q10: Do not catch every `AssertionError` from the stock test.

This block hides NaNs, out-of-bounds canaries, nondeterminism, and wrong output shapes. Limit the expected failure to the final numerical comparison. Use the actual KV length and a strict expected-failure rule.

test/attention/fmha/test_mem_eff_attention.py
```diff
@@ -155,9 +155,16 @@ def test_logsumexp(opFW_device_dtype_biasT_B_Mq_Mkv_H_K_Kv):
...
-    if op is fmha.ck.FwOp:
-        # relax numerical tolerance for CK FwOp
-        assert_allclose(lse, ref_lse, atol=2e-4, rtol=2e-4)
+    if op in (fmha.ck.FwOp, fmha.flydsl_forward.FwOp):
+        # MFMA kernels: large LSE * reduced-precision accumulation needs an rtol too.
+        if op is fmha.flydsl_forward.FwOp and max(q_len, kv_len) >= 256:
+            # At scale=3, kv_len >= 256 hits the f16/bf16 LSE floor (see test_forward).
+            try:
+                assert_allclose(lse, ref_lse, atol=2e-4, rtol=2e-4)
+            except AssertionError:
```
q10: Do not catch every `AssertionError` from the LSE test.

This block also hides shape and dtype errors. Limit the expected failure to the final numerical comparison. Use the actual KV length and a strict expected-failure rule.

mslk/attention/flydsl/flash_attn_utils.py
```diff
@@ -1314,9 +1351,15 @@ def _make_flash_attn_generic_traits(
...
-    if path_tag.upper() in ("N32", "N128"):
+    # Window / bias / dropout are N32-only (the N128 dual-subtile path does not
+    # gather per-subtile mask/bias), so force N32 when any is active.
+    _windowed = window_left >= 0
+    if _windowed or has_bias or has_dropout:
+        path = "N32"
+    elif path_tag.upper() in ("N32", "N128"):
         path = path_tag.upper()
-    elif dtype_str in ("f16", "bf16") and causal and head_dim == 128:
+    elif dtype_str in ("f16", "bf16") and head_dim == 128:
```
q10, line 1389: Clarify this scope change in the PR summary, or move it to a separate PR.

Removing the `causal` condition changes kernel selection for existing non-causal D=128 callers. This change is not limited to the new forward operator.

mslk/attention/fmha/flydsl_forward.py
avbokovoy: Fixed in https://github.com/meta-pytorch/MSLK/pull/485/commits/11e157157153eb7d5b51e65a76d81a97daf60c06 by setting the value to True

mslk/attention/fmha/flydsl_forward.py
avbokovoy: Reworked dropout in https://github.com/meta-pytorch/MSLK/pull/485/commits/ea75b638415b3f2f3456a7836fc6dbac1ad2020e, added test. The is that current FlyDSL backward implementation doesn't support dropout and to make FlyDSL forward compatible with ck.BwOp, we need to reuse CK generator, which will break `PYTHONONLY` build type. Documented that and suggest that dropout e2e dropout issues is to be resolved during backward dropout implementation.

mslk/attention/fmha/flydsl_forward.py
avbokovoy: Switched to uin8 type in https://github.com/meta-pytorch/MSLK/pull/485/commits/ea75b638415b3f2f3456a7836fc6dbac1ad2020e. See the comment in https://github.com/meta-pytorch/MSLK/pull/485#discussion_r3803058635

mslk/attention/fmha/flydsl_forward.py
avbokovoy: Done in https://github.com/meta-pytorch/MSLK/pull/485/commits/f9776db66adccff823be030d81c8e24b8362ae22

mslk/attention/flydsl/flash_attn_utils.py
avbokovoy: Implemented with buffer resource in https://github.com/meta-pytorch/MSLK/pull/485/commits/3ec93ef0c74636fd7ee5739962695889c2056dde

mslk/attention/flydsl/flash_attn_generic.py
avbokovoy: Done in https://github.com/meta-pytorch/MSLK/pull/485/commits/c31dba21ccb3b0f049ee44f1a3d2c9fe1a99bc2c

mslk/attention/flydsl/flash_attn_interface.py
avbokovoy: Done in https://github.com/meta-pytorch/MSLK/pull/485/commits/f41b50ea636d1a9b9711131ddd505c3bf10e4b30

mslk/attention/fmha/flydsl_forward.py
avbokovoy: Done in https://github.com/meta-pytorch/MSLK/pull/485/commits/68e9e6827d6d9bd35af6562781e8e047c0be709c

mslk/attention/flydsl/flash_attn_utils.py
avbokovoy: Done in https://github.com/meta-pytorch/MSLK/pull/485/commits/4fbc7ac7422bef8103235deaff7bfe587d985d6c

test/attention/fmha/test_forward_flydsl.py
avbokovoy: Done in https://github.com/meta-pytorch/MSLK/pull/485/commits/dc30ae180693046207f7f316ae1ab361d593b3e4

test/attention/fmha/test_mem_eff_attention.py
avbokovoy: Done in https://github.com/meta-pytorch/MSLK/pull/485/commits/b1d08ad12b789f9aedbd98f54eabbc94fdbd44ed

mslk/attention/flydsl/flash_attn_utils.py
avbokovoy: Documented in https://github.com/meta-pytorch/MSLK/pull/485/commits/2b6a0b4ece45c783562ff613447f0585cf7b7c4a, updated PR description

avbokovoy:
> This review contains the OSS-applicable findings from Meta review D115895891.
> 
> Please also address these test and scope items:
> 
>     * Use more than one head in dropout tests. Test f16 and bf16.
> 
>     * Add the 222-case parity sweep and benchmark source, or link reproducible results.
> 
>     * Link a task for the dualwave f16 and cross-attention NaN defects. The new exclusions move existing users to the generic path.
> 
> 
> The window mask, gappy page lookup, split-K layout, BMGHK folding, and cache key are correct. Meta-only BUCK and internal CI findings are intentionally omitted.

@q10 Thanks for the feedback. Regarding the points:
>  Use more than one head in dropout tests. Test f16 and bf16.
Done in https://github.com/meta-pytorch/MSLK/pull/485/commits/2f8124fefa9e9fe31d5dd695e174f89f87e3e57f

> Add the 222-case parity sweep and benchmark source, or link reproducible results.

Results might be reproduced by running:
```bash
python -m mslk.bench.attn.flydsl_fwd_bench --full
```

> Link a task for the dualwave f16 and cross-attention NaN defects. The new exclusions move existing users to the generic path.

Just let me know if we should work on that. I can't really "link a task"

Also, the PR was rebased and restructured to consolidate with merged FlyDSL backward file structure. Fixed an issue with JIT cache collision in BWD in https://github.com/meta-pytorch/MSLK/pull/485/commits/985d188a6bba45476fd46c75c232691de55657ea

Let me know if you have any questions

mslk/attention/fmha/__init__.py
```diff
@@ -59,6 +59,8 @@
 MemoryEfficientAttentionCkOp = (ck.FwOp, ck.BwOp)
 MemoryEfficientAttentionCkDecoderOp = (ck_decoder.FwOp, ck.BwOp)
 MemoryEfficientAttentionSplitKCkOp = (ck_splitk.FwOp, ck.BwOp)
+# FlyDSL forward op. Backward falls back to the CK backward (no FlyDSL bw yet).
```
bottler: Remove this comment, it isn't true.

mslk/attention/fmha/__init__.py
```diff
@@ -961,6 +963,7 @@ def merge_attentions(  # noqa: C901
     flash_mtia.FwOp,
     flash3.FwOp,
     triton_splitk.FwOp,
+    flydsl.FwOp,
```
bottler, line 964: nit: keep this alphabetical?

mslk/attention/fmha/dispatch.py
```diff
@@ -132,6 +132,9 @@ def _dispatch_fw_priority_list(
         priority_list_ops = deque(
             [
                 ck.FwOp,
+                # After CK so it does not displace the default, but available for
+                # inputs CK declines or when explicitly selected.
```
bottler: comment not needed

mslk/attention/fmha/flydsl.py
```diff
@@ -840,3 +858,570 @@ def apply(cls, ctx: Context, inp: Inputs, grad: torch.Tensor) -> Gradients:
...
+        BlockDiagonalCausalWithOffsetGappyKeysMask,
+        BlockDiagonalCausalLocalAttentionMask,
+        BlockDiagonalCausalLocalAttentionFromBottomRightMask,
+        PagedBlockDiagonalPaddedKeysMask,
+        PagedBlockDiagonalCausalWithOffsetPaddedKeysMask,
+        PagedBlockDiagonalGappyKeysMask,
+    )
+
+    # Dropout (dense): reuses CK's philox mask and records [seed, offset] in
+    # ctx.rng_state so ck.BwOp reproduces it. See _apply_bmhk.
```
bottler: I don't think this comment is true.

mslk/attention/fmha/flydsl.py
```diff
@@ -840,3 +858,570 @@ def apply(cls, ctx: Context, inp: Inputs, grad: torch.Tensor) -> Gradients:
...
+    try:
+        from mslk.flydsl.common import is_flydsl_available
+
+        return is_flydsl_available()
+    except Exception:
+        return False
+
+
+@register_operator
+class FwOp(AttentionFwOpBase):
```
bottler: If it would be easy, I think it'd be nice to put this before BwOp, because other files which define Fw and Bw do it in that order, and readers might prefer to think about Fw before Bw.

test/attention/fmha/test_flydsl_parity.py
```diff
@@ -0,0 +1,137 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""FlyDSL-vs-CK parity sweep and exported forward/backward end-to-end test.
+
+This is the reproducible source behind the PR's parity claim: it force-compares
```
bottler: Is this file meant to be committed? The phrase "the PR" is odd.

test/attention/fmha/test_forward_flydsl.py
```diff
@@ -0,0 +1,53 @@
...
+_ARGVALUES = [(*v, False, "BMHK") for v in _gen["argvalues"]]
+_IDS = [i + "-BMHK" for i in _gen["ids"]]
+
+
+@pytest.mark.parametrize(
+    "opFW_device_dtype_biasT_B_Mq_Mkv_H_K_Kv_packed_fmt",
+    _ARGVALUES,
+    ids=_IDS,
+)
+def test_forward_flydsl(opFW_device_dtype_biasT_B_Mq_Mkv_H_K_Kv_packed_fmt):
```
bottler, line 32: Any tests which are meant to be committed should have appropriate hardware annotations. Prefer filtering the parametrization to skipping in the test. We try to avoid any tests skipping on internal CI, see case_generation.py and logic around UNSUPPORTED_HARDWARE_EXPUNGE for examples.

mslk/attention/fmha/__init__.py
avbokovoy: Done in https://github.com/meta-pytorch/MSLK/pull/485/commits/89b962396989c852b0169740e3dc73a2abd40b0a

mslk/attention/fmha/__init__.py
avbokovoy: Done in https://github.com/meta-pytorch/MSLK/pull/485/changes/3a03073c081f71ecd4c828d6bcd93b205f4c09cc

mslk/attention/fmha/dispatch.py
avbokovoy: Removed in https://github.com/meta-pytorch/MSLK/pull/485/changes/53d35ab8c9d4717c342f4e6abc47a9262c736115

mslk/attention/fmha/flydsl.py
avbokovoy: Indeed, after recent changes it's outdated. Removed in https://github.com/meta-pytorch/MSLK/pull/485/changes/946cc9173b5d432822b8e2365d177ab67fbacdb6

mslk/attention/fmha/flydsl.py
avbokovoy: Sure, done in https://github.com/meta-pytorch/MSLK/pull/485/commits/338526ab71cf3ad991c15ef096f1d7698810469c

test/attention/fmha/test_flydsl_parity.py
avbokovoy: Let's just remove the whole file for simplicity. I believe current tests are sufficient for parity check. Done in https://github.com/meta-pytorch/MSLK/pull/485/commits/52f63f725043d864dfd09f15a9d16e90854fda89

test/attention/fmha/test_forward_flydsl.py
avbokovoy: Done in https://github.com/meta-pytorch/MSLK/pull/485/commits/0546ef946d01d9926a40c840c73757010f536c49. Could you please have a look?

bottler: If it's easy, I'd prefer actually filtering _ARGVALUES, at least for things which are easily predictable. But this is okay.

bottler (approved):

PR #495, opened 2026-08-21
Add FP8×INT4 rowwise GEMM for ROCm (f8i4bf16_rowwise)

liligwu:
This PR is a copy of https://github.com/apicciau/MSLK/pull/3

## Summary

Adds a Triton implementation of `f8i4bf16_rowwise` (FP8 activations × INT4 weights → BF16 output) for AMD MI300X/MI350X. The kernel reuses the unified `_bf16i4_rowwise_kernel` from the BF16×INT4 path, gated on a compile-time `HAS_X_SCALE` constexpr — no kernel fork.

## Context

- Hardware: AMD Instinct MI300X (gfx942), MI350X (gfx950, CDNA4)
- Depends on #1 (`rocm/bf16i4-shuffled-grouped`) — this PR stacks on top of that branch
- Upstream PR: meta-pytorch/MSLK#479

## Implementation

**Shared kernel with HAS_X_SCALE constexpr.** The existing `_bf16i4_rowwise_kernel` gains two new parameters: `x_scale_ptr` (per-row FP8 activation dequant scale) and `HAS_X_SCALE: tl.constexpr = False`. When `HAS_X_SCALE=False` (BF16 path), the kernel behaves identically to before — the default value preserves backward compatibility and the dummy pointer is never dereferenced.

**Deferred BF16 upcast.** The `.to(tl.bfloat16)` cast on activation loads is moved from the load site to a single centralised location after the four masking branches. This is required because FP8 inputs return a native FP8 tensor type from `tl.load` that must be upcast explicitly — but for BF16 inputs the cast is a no-op, so both paths share the same code.

**Two-dot path for FP8.** When `HAS_X_SCALE=True`, the kernel issues two separate `tl.dot` calls (one for even-K, one for odd-K) instead of fusing via `tl.cat`. The `tl.cat` fusion concatenates tiles in LDS before the dot, which causes bank conflicts that are a net loss for FP8 activations on MI300X. Two separate dots avoid the LDS staging at the cost of two MFMA instructions — a net win for this dtype. The BF16 path retains the fused `tl.cat` dot (controlled by `FUSE_DOT`).

**Per-row activation scale epilogue.** After the K-loop, the float32 accumulator is multiplied element-wise by `x_scale[m]` before storing to the workspace. This fuses the FP8 dequantisation scale into the GEMM epilogue rather than requiring a separate pointwise kernel.

**FP8 autotuner pruning.** Four additional prune rules fire only when `HAS_X_SCALE=True`: skip `BLOCK_N < 128` for large N, skip `BLOCK_K > 64`, skip `SPLIT_K == 2`, fix `GROUP_SIZE_M = 4`. These were determined by profiling winning configs on MI350X and reduce the FP8 autotuning search space by ~5×.

**Thin Python wrapper.** `f8i4bf16_rowwise_gemm.py` byte-splits FP8 activations into even/odd K columns, reinterprets them as FP8 `TensorWrapper`s via `reinterpret_fp8_type`, and calls the shared kernel with `HAS_X_SCALE=True`. The wrapper also coerces scale/zp tensors to float32. Registered as `mslk::f8i4bf16_rowwise` on ROCm via `torch.library.impl`.

## Testing

All 9 existing BF16×INT4 accuracy tests pass (no regressions):

```
9 passed in 74s
```

FP8×INT4 accuracy against dequantised reference (`deq(FP8_x) @ deq(INT4_w)`):

| Shape | M | N | K | max error | mean error |
|---|---:|---:|---:|---:|---:|
| Small | 1 | 256 | 1024 | 0.0005 | 0.0001 |
| Medium | 16 | 2048 | 1024 | 0.0020 | 0.0001 |
| Large | 512 | 4096 | 2048 | 0.0039 | 0.0002 |

FP8×INT4 dispatch test (`torch.ops.mslk.f8i4bf16_rowwise` vs direct call): bitwise identical on all shapes.

## Performance

Hardware: AMD Instinct MI350X (gfx950), ROCm 7.1.1, Triton 3.5.1+rocm7.1.1.
`triton.testing.do_bench`, warmup=25, rep=200.
Peak: 4600 TFLOPS FP8, 2300 TFLOPS BF16, 8000 GB/s HBM.

### f8i4bf16_rowwise vs bf16i4bf16_rowwise — decode shapes (memory-bandwidth bound)

Model-attributed shapes. FP8 and BF16 paths are both kernel-launch-latency bound at small M; performance converges.

| Model | M | N | K | FP8 (μs) | BF16 (μs) | FP8 TFLOPS | BF16 TFLOPS |
|---|---:|---:|---:|---:|---:|---:|---:|
| llama4 | 1 | 896 | 5120 | 46.3 | 44.8 | 0.2 | 0.2 |
| llama4 | 16 | 896 | 5120 | 45.4 | 44.8 | 3.2 | 3.3 |
| llama4 | 64 | 896 | 5120 | 45.7 | 44.3 | 12.9 | 13.2 |
| llama4 | 128 | 2048 | 5120 | 46.1 | 44.3 | 58.2 | 60.6 |
| llama3_70b | 1 | 8192 | 3584 | 40.6 | 38.8 | 1.4 | 1.5 |
| llama3_70b | 16 | 8192 | 3584 | 40.0 | 38.3 | 23.5 | 24.5 |
| llama3_70b | 64 | 8192 | 3584 | 41.5 | 40.7 | 90.6 | 92.4 |
| llama3_70b | 128 | 7168 | 8192 | 84.6 | 87.9 | 177.7 | 171.0 |

### f8i4bf16_rowwise vs bf16i4bf16_rowwise — prefill shapes (compute bound)

FP8 MFMA has 2× the throughput of BF16 MFMA. The FP8 path is 5–8% faster at large M.

| Model | M | N | K | FP8 (μs) | BF16 (μs) | FP8 TFLOPS | % FP8 peak | BF16 TFLOPS | % BF16 peak |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| llama3_70b | 2048 | 8192 | 3584 | 210.6 | 218.2 | 571.1 | 12.4% | 551.1 | 24.0% |
| llama3_70b | 2048 | 7168 | 8192 | 396.6 | 428.2 | 606.4 | 13.2% | 561.7 | 24.4% |
| llama3_70b | 4096 | 7168 | 8192 | 768.4 | 826.0 | 626.1 | 13.6% | 582.4 | 25.3% |

### DIFF
--- a/bench/gemm/gemm_ops.py
+++ b/bench/gemm/gemm_ops.py
@@ -11,24 +11,9 @@ from enum import auto, Enum
 import torch
 
 import mslk.gemm  # noqa: F401 — ensure mslk namespace exists
-
-_STUB_SCHEMAS = [
-    ("f8i4bf16_rowwise", "(Tensor XQ, Tensor WQ, Tensor x_scale, Tensor w_scale, Tensor w_zp) -> Tensor"),
-]
-for _name, _schema in _STUB_SCHEMAS:
-    if not hasattr(torch.ops, "mslk") or not hasattr(torch.ops.mslk, _name):
-        torch.library.define(f"mslk::{_name}", _schema)
-
 from mslk.bench.common.utils import BenchOptions, do_bench
 from mslk.flydsl.common import is_flydsl_available
 from mslk.gemm.triton.fp8_gemm import matmul_fp8_block, matmul_fp8_row, to_mxfp8
-
-if is_flydsl_available():
-    from mslk.gemm.flydsl.preshuffle_gemm import (
-        flydsl_preshuffle,
-        flydsl_preshuffle_batched_gemm,
-        flydsl_preshuffle_gemm,
-    )
 from mslk.gemm.triton.grouped_gemm import grouped_gemm, grouped_gemm_fp8_rowwise
 from mslk.quantize.shuffle import (
     ck_preshuffle,
@@ -51,6 +36,20 @@ from mslk.quantize.triton.fp8_quantize import (
     quantize_fp8_tensor,
 )
 from mslk.utils.device import is_cuda, is_gfx942, is_gfx950, is_rocm
+
+if is_flydsl_available():
+    from mslk.gemm.flydsl.preshuffle_gemm import (  # noqa: E402
+        flydsl_preshuffle,
+        flydsl_preshuffle_batched_gemm,
+        flydsl_preshuffle_gemm,
+    )
+
+_STUB_SCHEMAS = [
+    ("f8i4bf16_rowwise", "(Tensor XQ, Tensor WQ, Tensor x_scale, Tensor w_scale, Tensor w_zp) -> Tensor"),
+]
+for _name, _schema in _STUB_SCHEMAS:
+    if not hasattr(torch.ops, "mslk") or not hasattr(torch.ops.mslk, _name):
+        torch.library.define(f"mslk::{_name}", _schema)
 
 try:
     from tinygemm.utils import group_quantize_tensor
--- a/bench/gemm/gemm_ops.py
+++ b/bench/gemm/gemm_ops.py
@@ -9,6 +9,16 @@ import functools
 from enum import auto, Enum
 
 import torch
+
+import mslk.gemm  # noqa: F401 — ensure mslk namespace exists
+
+_STUB_SCHEMAS = [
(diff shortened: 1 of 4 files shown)


q10:
Merged in https://github.com/meta-pytorch/MSLK/pull/502

PR #497, opened 2026-08-22
FlyDSL forward attention op (#497)

q10:
Summary:

Adds `flydsl_forward.FwOp`, a FlyDSL flash-attention forward operator that wraps `mslk.attention.flydsl.flydsl_flash_attn_func` behind the `AttentionFwOpBase` interface as a drop-in for `ck.FwOp`, together with the kernel-side features it needs (split-K, sliding-window tile skipping, in-kernel gather for gappy/padded/paged KV, dropout, and LSE on the generic path).

- **dtypes / head dim**: bf16 and f16; any `head_dim` in `[1, 256]` (dims the kernel does not build natively are zero-padded to the next of `{64, 128, 256}`, with the native softmax scale pinned so padding does not change the result).
- **formats**: BMHK and BMGHK (GQA); the G axis is folded into H the same way `ck.FwOp` does.
- **attn_bias**: full parity with CK, all 18 xformers types -- causal / non-causal self-attention, additive tensor bias, sliding-window (local) masks, varlen (`BlockDiagonal*`), padded/gappy keys, and paged KV. Paged KV accepts any page_size that is a multiple of 64 (remapped to 64-row sub-pages); paged-gappy uses an in-kernel per-row page lookup.
- **LSE** output supported (including the generic paged path).
- **dropout** matches CK's philox exactly (via `torch.ops.xformers._ck_rand_uniform`) and records `[seed, offset]` in `ctx.rng_state` so `ck.BwOp` reproduces the mask.
- **split-K** on the generic kernel for short-Q dense/causal shapes, gated by a Python port of CK's occupancy heuristic; a combine kernel reduces the per-split partials into the final O and LSE.
- **CUDA-graph**: every supported bias path is graph-capturable (no device-to-host syncs on the capture path).
- Registered in `ALL_FW_OPS` and the ROCm dispatch priority list after `ck.FwOp`: it does not displace the default (auto-dispatch still selects `ckF`), but is available when CK declines or when the op is selected explicitly.
- Backward falls back to `ck.BwOp` (no FlyDSL backward).

## Behavior change / scope (not limited to the new op)

  - N128 kernel selection now applies to non-causal D=128 (f16/bf16), not just causal. Previously the N128 generic path was gated on causal; that gate was removed, so existing non-causal D=128 callers of the
  generic kernel now select N128 instead of N32. N128 is faster for D=128, but this changes the kernel used for callers unrelated to flydslF. (See flash_attn_utils.py:_make_flash_attn_generic_traits.)

## Correctness

A full sweep runs `flydsl_forward.FwOp` and `ck.FwOp` on identical inputs across the whole CK-supported matrix (all 18 attn_bias types, bf16 and f16, BMHK and BMGHK/GQA, head dims 64/96/128/256, short and long q, varied batch), comparing wherever each op accepts the case. Result: 222 cases, 0 failures, 0 skips. Every case matches CK within the reduced-precision floor (worst bf16 relerr 0.019, f16 far lower), with no NaN. Zero skips means the op accepts every case CK accepts.

## Performance

CUDA-graph replay timing on gfx950 (MI355X), FlyDSL vs `ck.FwOp` (the op this replaces). `fly/ck < 1` means FlyDSL is faster. Inputs use the test suite's scale-3 tensors, so error is reported relative to CK output magnitude (bf16/f16 floor is about 0.004 to 0.02). Full per-shape tables in `BENCHMARKS.md`.

Representative bf16 results:

| bias type        | shape (B,q,kv,D)   | FlyDSL us | CK us  | fly/ck |
|------------------|--------------------|-----------|--------|--------|
| none             | 1, 4096, 4096, 128 | 126.0     | 234.0  | 0.54   |
| causal           | 1, 4096, 4096, 128 | 111.0     | 216.9  | 0.51   |
| window           | 1, 2048, 2048, 128 | 13.4      | 22.7   | 0.59   |
| varlen_causal    | 8, 1024, 1024, 128 | 113.7     | 227.1  | 0.50   |
| varlen_causal_br | 4, 512, 512, 128   | 36.6      | 84.6   | 0.43   |
| gappy_causal     | 4, 256, 512, 128   | 43.9      | 83.8   | 0.52   |
| paged            | 4, 256, 512, 128   | 37.5      | 47.2   | 0.79   |
| paged_gappy      | 4, 256, 512, 128   | 42.7      | 51.0   | 0.84   |
| tensorbias       | 1, 2048, 2048, 128 | 114.1     | 106.1  | 1.08   |

Representative f16 results (same shapes):

| bias type        | shape (B,q,kv,D)   | FlyDSL us | CK us  | fly/ck |
|------------------|--------------------|-----------|--------|--------|
| none             | 1, 4096, 4096, 128 | 127.0     | 260.2  | 0.49   |
| causal           | 1, 4096, 4096, 128 | 105.3     | 209.2  | 0.50   |
| window           | 1, 2048, 2048, 128 | 15.2      | 25.0   | 0.61   |
| varlen_causal    | 8, 1024, 1024, 128 | 158.3     | 211.2  | 0.75   |
| varlen_causal_br | 4, 512, 512, 128   | 32.0      | 70.8   | 0.45   |
| gappy_causal     | 4, 256, 512, 128   | 35.2      | 62.7   | 0.56   |
| paged            | 4, 256, 512, 128   | 35.9      | 40.3   | 0.89   |
| paged_gappy      | 4, 256, 512, 128   | 42.6      | 49.0   | 0.87   |
| tensorbias       | 1, 2048, 2048, 128 | 121.7     | 110.7  | 1.10   |

FlyDSL is faster on the large majority of type/shape/format combinations in both dtypes, with the biggest wins on long-Q prefill (up to ~2x at S=4096), sliding-window at large S (~1.6x), and cross-length varlen causal (up to 2x). f16 tracks bf16 closely and is numerically tighter (relerr ~0.0005 to 0.0006 vs CK, since it skips the bf16 mantissa floor). GQA tracks BMHK. The few near-parity cases are consistent across both dtypes: additive tensor bias at S=2048 (intrinsic full-plane bias streaming, ~1.05 to 1.10x), small-batch varlen_causal (launch-bound, ~1.10x), and paged GQA (~1.07 to 1.09x).

## Testing

- `test/attention/fmha/test_forward_flydsl.py` -- stock forward suite scoped to the op (skips declined cases; xfails the documented scale-3 multi-tile precision floor).
- `test/attention/fmha/test_dropout_flydsl.py` -- philox parity, reproducibility, and forward+backward vs CK.
- `test/attention/fmha/test_mem_eff_attention.py` -- LSE and forward coverage for the op (exercised via `ALL_FW_OPS`, including GQA and all bias types).

## Known limitations

- No FlyDSL backward; backward uses `ck.BwOp`.

Reviewed By: bottler

Differential Revision: D115895891

Pulled By: q10

### DIFF
--- a/bench/attn/flydsl_fwd_bench.py
+++ b/bench/attn/flydsl_fwd_bench.py
@@ -1,0 +1,314 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""FlyDSL-vs-CK forward FMHA sweep (correctness + latency).
+
+Runs ``flydsl.FwOp`` and ``ck.FwOp`` on identical inputs across the
+CK-supported bias-type matrix (built with the test suite's ``create_tensors`` so
+the varlen / paged / gappy layouts are exactly what the tests use), timing both
+ops and computing the relative error wherever both accept the case. This is the
+reproducible source for the PR's parity + representative-latency tables: the
+representative rows are simply filtered from the full sweep and re-printed at the
+end.
+
+Bias categories (the labels used in the representative tables):
+    none causal window varlen_causal varlen_causal_br gappy_causal
+    paged paged_gappy tensorbias
+
+Examples::
+
+    # Representative rows only (fast) -> reproduces the two tables:
+    python -m mslk.bench.attn.flydsl_fwd_bench
+
+    # Full sweep across head dims 64/96/128/256 (approaches the ~222-case sweep):
+    python -m mslk.bench.attn.flydsl_fwd_bench --full --head-dims 64,96,128,256
+"""
+
+import os
+import sys
+from dataclasses import dataclass
+from typing import Dict, List, Optional, Tuple
+
+import click
+import torch
+from mslk.attention import fmha
+from mslk.attention.fmha import flydsl
+
+
+def _load_create_tensors():
+    """Import the test suite's create_tensors (handles every bias layout)."""
+    rel = os.path.join("test", "attention", "fmha", "case_generation.py")
+    cands = [os.path.abspath(os.path.join(os.path.dirname(__file__), "..", ".."))]
+    d = os.getcwd()
+    for _ in range(8):
+        cands.append(d)
+        d = os.path.dirname(d)
+    root = next((c for c in cands if os.path.exists(os.path.join(c, rel))), None)
+    if root is None:
+        raise SystemExit(
+            "Could not locate test/attention/fmha/case_generation.py; run from the "
+            "MSLK repo root."
+        )
+    sys.path.insert(0, os.path.join(root, "test", "attention"))
+    try:
+        from fmha.case_generation import create_tensors  # noqa: E402
+    except Exception as e:  # pragma: no cover - dev tool
+        raise SystemExit(f"Failed importing create_tensors ({e}); pytest installed?")
+    return create_tensors
+
+
+_ab = fmha.attn_bias
+# label -> attn_bias type passed to create_tensors
+BIAS_CASES: Dict[str, object] = {
+    "none": type(None),
+    "causal": _ab.LowerTriangularMask,
+    "window": _ab.LowerTriangularFromBottomRightLocalAttentionMask,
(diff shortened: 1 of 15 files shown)


bottler (approved):

